PR #22, opened 2026-06-06
One fast parse path: retain source instead of a second document/locator

samtalki:
Supersedes the earlier data-only-path approach on this branch, per review.

## Why
The review asked two things: is the data-only path over-optimizing, and would the document/locator logic be duplicated for future formats. Both land on the same answer — collapse the dual path.

- The data-only path had **no shipping consumer**: the CLI, TUI, and Python bindings all used the lossless path, and the win on real cases was tens of µs.
- A converter re-serializes to a *different* format from the typed `MpcCase`, so it never needs the MATPOWER source document. The dual-function split was the wrong shape.

## What
- `parse_matpower` is now the single parse path: it retains the original source text (`Arc<str>`) and `write_matpower` echoes it, so the byte-exact round-trip costs no extra pass. `parse_matpower_data` / `parse_matpower_data_file` are removed.
- `parser/matpower/document.rs` → `locate.rs`: `MatpowerDocument` / `build_document` are gone. There is now **one** assignment locator (`locate_assignments`), not two algorithms kept byte-identical by a vendored-case test.
- `timeparse`, the benchmark, and the README/RESULTS reflect the single path. Re-measured: no slower than the old data-only column (e.g. case193k 158 ms).

## Tests
Round-trip + typed-data tests stay green across the vendored cases; the unbalanced-matrix guard and source-presence behaviour are covered in `parser/matpower/tests.rs`.


cameronkhanpour:
Is there any worry of over optimizing here? Do we need to duplicate this logic once we want to extend to egret json files and PSS/E files?

samtalki:
@cameronkhanpour good call on both — acted on it.

**Over-optimizing:** yes. The data-only path had no shipping consumer (CLI/TUI/Python bindings all used the lossless path) and the real-case win was tens of µs. I dropped `parse_matpower_data` and made the single `parse_matpower` fast by default: it retains the source text and the writer echoes it, so the byte-exact round-trip costs no extra pass. Measured no slower than the old data-only column.

**Duplication for EGRET/PSS-E:** the data/document split was MATPOWER-text-specific (it keys on `mpc.<field> = [...]`), so it wouldn't have transferred to JSON (serde) or PSS-E (fixed-format) regardless — `MpcCase` is the shared layer every format lands on. But you were right that it *already* duplicated a locator inside MATPOWER: `build_document` and `locate_assignments` were two algorithms that had to yield byte-identical spans, kept in sync only by a vendored-case test. That's now one locator (`locate.rs`).

The actual converter work (PowerModels/EGRET JSON over a format-neutral hub, plus readers) is going up as a separate PR so this one stays a focused cleanup.

Issue #55, opened 2026-06-08
Roadmap: ML data-layer and polyglot embeddability follow-ups


samtalki:
Tracking issue for the next round of work that deepens powerio's two core strengths: the embeddable ML/matrix data layer and multi-language reach over the `Network` hub. The gridfm scenario-batch **export** shipped in #57 (`write_gridfm_batch`, multiple `gridfm` CLI inputs, and the Python `gridfm` extra). The topology-invariant **operator** reuse that #14 actually specifies (build the operators once and reuse the DC Cholesky across scenarios) is a separate, still-open item — see #14. This issue collects the remaining work so it can be sequenced independently.

## Polyglot embeddability
- [ ] **gridfm-schema Arrow C Data Interface export over the C ABI (#38).** Make the gridfm `RecordBatch`es reachable from `powerio-capi` (depend on `powerio-matrix`, or share the batch builders) so Julia/C consumers get the `bus_data`/`gen_data`/`branch_data`/`y_bus_data` tables zero-copy, not just the lighter raw-network Arrow schema that ships today.
- [ ] **PowerIO.jl over the C ABI.** Register the package and add native bridges (PowerModels.jl, ExaModelsPower.jl, PowerDiff.jl).
- [ ] **Zero-copy Arrow handoff into Python.** Hand the gridfm tables to Python via pyarrow (C Data Interface) so callers can skip the Parquet round-trip; pairs with #38.

## ML / matrix depth
- [ ] **Operator reuse across injection-only scenario batches (#14).** Build the topology-invariant operators once and reuse a single grounded-Laplacian Cholesky for all per-scenario DC solves. Valid only when status/shunts are fixed (injection-only) — see the correction on #14.
- [ ] **Distinct `scenario` vs `load_scenario_idx` in the gridfm export.** powerio currently stamps both columns with the same per-snapshot id, so graphkit's train/val/test stratification (which groups on `load_scenario_idx`) treats every snapshot as its own stratum. Let a caller set `load_scenario_idx` independently — e.g. group several topology variants under one base load scenario — to match datakit's two-level (load scenario × topology variant) keying.
- [ ] **Sparse PTDF for large grids.** PTDF is currently dense `m×n` via the self-contained Cholesky in `matrix/sensitivity.rs`; add a sparse/iterative path for large networks.
- [ ] **LinDist3Flow matrices** over the petgraph spanning-tree view (`IndexedNetwork::to_petgraph`).

## Format breadth & fidelity
[...]

Issue #38, opened 2026-06-07
Matrix C ABI: CSR matrices and matrix-backed Arrow export


samtalki:
Follow-up to the C ABI keystone (`powerio-capi`, #37) and the roadmap C ABI (#8).

`powerio-capi` covers the `powerio` surface: parse / write / convert / query / numeric-table extraction. A C/Julia/C++ consumer can build matrices itself from the extracted tables (`from`, `to`, `1/x`, nodal demand/shunt, base_mva). The next step is to hand back the **assembled** sparse matrices over the same ABI style, so consumers don't re-implement the builders.

## Scope
A `powerio-matrix-capi` crate (cdylib + staticlib), depending on `powerio-matrix`, mirroring `powerio-capi`'s conventions (`cm_` prefix, panics caught at the boundary, owned handles/strings freed by the library).

Surface sketch:
- `cm_parse(path, from, err, errlen) -> *CmCase` / `cm_case_free`
- `cm_build(case, kind, err, errlen) -> *CmMatrix` where `kind` ∈ {`bprime`,`bdoubleprime`,`ybus_real`,`ybus_imag`,`lacpf`,`adjacency`,`ptdf`,`lodf`}
- `cm_matrix_rows/cols/nnz`, `cm_matrix_indptr/indices/data` (lend `const` pointers into library-owned CSR; caller copies), `cm_matrix_free`
- DC-OPF bundle: either `cm_write_dcopf_bundle(case, out_dir, ...)` (reuse `opf_pipeline`) or per-operator `cm_build` kinds.

## Notes
- CSR indices as `int32_t` (scipy/most BLAS) with a guard past `i32::MAX`, matching `powerio-py`'s COO narrowing.
- Header checked in (`powerio-matrix-capi/include/powerio-matrix.h`), cbindgen-regenerated.
- Add a Julia/C smoke test in CI once a toolchain step exists.

Refs #8, #37.

samtalki:
Cross-link: #52 (gridfm Parquet export) introduces `gridfm_record_batches`, which builds the network as Arrow `RecordBatch`es (bus/gen/branch/y_bus) and feeds the Parquet writer. That's the columnar complement to this crate's CSR matrices — same C ABI surface, just self-describing columnar tables instead of `indptr/indices/data`.

[...]

PR #58, opened 2026-06-08
feat(mcp): expose powerio as an MCP convert/validate tool (#26)

### LINKED ISSUE #26: Expose powerio as an MCP convert/validate tool
The Python binding already has \`powerio.convert(path, to, from_=None) -> Conversion\` and \`parse_matpower\`. Wrap these as an MCP server so LLM-agent tooling gets a fast, lossless \`convert\` and a \`validate\`/\`summary\` tool over the case-file family.

Agentic power-system stacks typically wire each underlying tool to its own format and have no unified converter; powerio fills that gap.

**Scope**
- A small MCP server (Python, over the \`powerio\` package) exposing:
  - \`convert_case(input, to, from=None) -> {text, warnings}\`
  - \`case_summary(path) -> {buses, branches, gens, loads, ...}\`
- Lives alongside the Python package; no new Rust needed.

Context: collaborators building agent tooling are expected to consume this.

qian-harvard:
Closes #26.

Wraps the existing `powerio` Python package as a small [MCP](https://modelcontextprotocol.io) (FastMCP) server, so LLM-agent tooling gets a fast, lossless converter and a case summary over the case-file family. **No new Rust** — it reuses `powerio.convert`/`parse`/`parse_str`.

## Tools (over stdio)

Both accept either a filesystem `path` or inline `content`:

- **`convert_case(to, path|content, from_)` → `{text, warnings}`** — convert between formats with fidelity warnings. Since `convert()` is path-only, inline content is staged to a temp file (removed in a `finally`); `from_` is required for inline input.
- **`case_summary(path|content, format)`** — name, base MVA, source format, element counts, and connectivity. Stays scipy/numpy-free.

`PowerIOError`/`FileNotFoundError` map to clean `ValueError`s (FastMCP turns these into clean MCP tool errors rather than a transport crash).

## Packaging

- New `python/powerio/mcp/` submodule (`server.py`, `__init__.py`, `__main__.py`) + a `powerio-mcp` console script.
- Gated behind a `powerio[mcp]` optional extra. The `mcp` SDK needs **Python ≥3.10**, so the extra carries a `python_version >= '3.10'` marker — the core stays 3.9-compatible (abi3-py39 wheel unaffected).
- The submodule is **never imported from `powerio/__init__.py`**, so `import powerio` stays zero-dependency (enforced by the existing subprocess test).

## Verification

- **Full Python suite: 46 passed, 1 skipped** (the skip is an unrelated `gridfm`-extra test), including the new `test_mcp.py` (7 tests: path + inline for both tools, input/error guards) and the zero-dep contract test.
- Confirmed `import powerio` pulls in no `mcp`/`numpy`/`scipy`.
- `powerio-mcp` console script installs and registers both tools.
- End-to-end stdio JSON-RPC handshake against the live `powerio-mcp` process returns a correct `case_summary` for `case9`.

### DIFF
--- a/.github/workflows/python.yml
+++ b/.github/workflows/python.yml
@@ -39,11 +39,13 @@ jobs:
           python-version: ${{ matrix.python }}
       - uses: dtolnay/rust-toolchain@stable
       - uses: Swatinem/rust-cache@v2
-      # `pip install .[all]` builds the extension via the maturin backend and
-      # pulls scipy/numpy/networkx so the matrix and graph tests run. Quote the
-      # extras so the shell doesn't glob `[all]`.
+      # `pip install .[all,mcp]` builds the extension via the maturin backend and
+      # pulls scipy/numpy/networkx (matrix + graph tests) plus the MCP SDK. The
+      # `mcp` extra carries a `python_version >= '3.10'` marker, so it installs
+      # the SDK on 3.12 (test_mcp runs) and resolves to nothing on 3.9 (test_mcp
+      # skips). Quote the extras so the shell doesn't glob `[all,mcp]`.
       - name: Install powerio + test deps
-        run: pip install '.[all]' pytest
+        run: pip install '.[all,mcp]' pytest
       - name: Run tests
         run: pytest python/tests -q
       # The zero-dep promise: parse/convert/write import nothing but the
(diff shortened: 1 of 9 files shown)


samtalki:
Would it be useful for you if I add zero-copy Arrow handoff into the Python interface for you @qian-harvard ? Would appreciate your thoughts on issues (#55) and (#38). I think there needs to be some sort of low level solution feeding into this mcp server, but I want to get it right the first time.

PR #69, opened 2026-06-09
Prepare release API and validation surfaces

samtalki:
## Summary

Prepare PowerIO for the public release branch:

- add the canonical Rust, Python, and C ABI API names (`parse_file`, `parse_str`, `convert_file`, `to_format`, `to_matpower`)
- bump the C ABI surface to v2 and regenerate `powerio.h`
- add Python stubs, docs, and wheel publishing workflow support for PyPI trusted publishing
- add the shared language API table and Python release notes
- prefer Polars for GridFM reads while keeping pandas/pyarrow as a compatibility extra
- replace the vulnerable `fast-float` dependency with `lexical-core`
- refresh validation and benchmark scripts for the canonical API

This PR does not publish to crates.io or PyPI. The PyPI workflow only runs from a GitHub release after release approval.

## Validation

- `cargo fmt --all --check`
- `cargo clippy --all-targets --all-features -- -D warnings`
- `cargo test`
- `cargo test -p powerio-capi --features arrow`
- `RUSTDOCFLAGS=-Dwarnings cargo doc --workspace --no-deps`
- `cargo audit`
- `maturin develop --release -E all -E gridfm -E pandas -E mcp`
- `pytest python/tests -q`
- `pdoc -o /private/tmp/powerio-pdoc python/powerio`
- `maturin build --release` and `maturin sdist`, followed by `twine check`
- `pip-audit` for base package and optional extras
- `bash benchmarks/run_validation.sh`
- Julia parser benchmark and Python parser benchmark refresh
- PowerIO.jl package tests against the local C ABI
- PowerDiff parser parity smoke against the local PowerIO C ABI


powerio-matrix/src/io/gridfm.rs
```diff
@@ -311,6 +305,86 @@ fn snapshot_views<'a>(snapshots: &'a [GridfmSnapshot<'a>]) -> Result<Vec<Snapsho
...
+        finite(scenario, "branch", row, "tap", br.tap)?;
+        finite(scenario, "branch", row, "shift", br.shift)?;
+        finite(scenario, "branch", row, "angmin", br.angmin)?;
+        finite(scenario, "branch", row, "angmax", br.angmax)?;
+        finite(scenario, "branch", row, "rate_a", br.rate_a)?;
+    }
+    for (row, g) in net.generators.iter().enumerate() {
+        finite(scenario, "generator", row, "pg", g.pg)?;
+        finite(scenario, "generator", row, "qg", g.qg)?;
+        finite(scenario, "generator", row, "pmax", g.pmax)?;
```
qian-harvard: **🔴 Correctness — gridfm export now rejects legitimate `±Inf` limit fields (regression).**

`finite()` uses `value.is_finite()`, which is `false` for `±Inf`. But `±Inf` is a valid *unbounded* sentinel here: `powermodels.rs:587-590` defaults absent `pmax/pmin/qmax/qmin` to `±f64::INFINITY` (reactive limits are routinely omitted), and `convert.rs:382` (`powermodels_unbounded_limit_round_trips_as_infinity`) codifies that `qmax=Inf` must survive as `+Inf`. MATPOWER `Inf` tokens parse the same way.

Same issue on branch `angmin/angmax` (lines 337-338) and bus `vmax/vmin` (lines 320-321). Net effect: `powerio gridfm <PowerModels/PGLib case with an unbounded limit>` now aborts with `Error::NonFiniteGridfmValue`, though it exported fine before this PR (Parquet f64 stores `±Inf` natively).

**Fix:** reject only `NaN` on limit-type fields (keep the `is_finite` check for `vm/va/p/q/r/x/...`). Secondary: this loop validates *out-of-service* generators too, which the export merely records as `in_service=0` — consider skipping them.

powerio-matrix/src/io/gridfm.rs
```diff
@@ -311,6 +305,86 @@ fn snapshot_views<'a>(snapshots: &'a [GridfmSnapshot<'a>]) -> Result<Vec<Snapsho
     Ok(views)
 }
 
+fn validate_snapshot_inputs(net: &Network, scenario: i64) -> Result<()> {
+    if net.is_normalized() {
+        return Err(Error::NormalizedGridfmSnapshot { scenario });
+    }
+    if !net.base_mva.is_finite() || net.base_mva <= 0.0 {
```
qian-harvard: **Cleanup (reuse).** This positive-finite `base_mva` guard + `Error::InvalidBaseMva` duplicates `normalize.rs:282-283` (`!(base.is_finite() && base > 0.0)`). The two copies must stay in lockstep with no compiler help. Consider a shared `Network::check_base_mva()` — it would sit naturally next to the `is_normalized()` call just above.

powerio-matrix/src/io/gridfm.rs
```diff
@@ -311,6 +305,86 @@ fn snapshot_views<'a>(snapshots: &'a [GridfmSnapshot<'a>]) -> Result<Vec<Snapsho
...
+    for (row, g) in net.generators.iter().enumerate() {
+        finite(scenario, "generator", row, "pg", g.pg)?;
+        finite(scenario, "generator", row, "qg", g.qg)?;
+        finite(scenario, "generator", row, "pmax", g.pmax)?;
+        finite(scenario, "generator", row, "pmin", g.pmin)?;
+        finite(scenario, "generator", row, "qmax", g.qmax)?;
+        finite(scenario, "generator", row, "qmin", g.qmin)?;
+        if let Some(cost) = &g.cost {
+            if cost_representable(Some(cost)) {
+                for (k, &coeff) in cost.coeffs.iter().take(cost.ncost).enumerate() {
```
qian-harvard: **Cleanup (reuse).** `coeffs.iter().take(cost.ncost)` + `cost_representable` re-encodes which coefficients `gridfm_cost` (~line 846) actually writes to the `cp0/cp1/cp2` columns — a third independent copy of that selection. Validate the exact tuple instead: `let (c0,c1,c2) = gridfm_cost(g.cost.as_ref()); finite(.., "cp0", c0)?; ...`. As-is the error names `coeffs[0]`, not the destination `cp*` column. Minor: lines 348-349 unwrap `Some(cost)` then immediately re-wrap as `Some(cost)` for `cost_representable` — collapse to `g.cost.as_ref().is_some_and(...)`.

.github/workflows/julia-binding.yml
```diff
@@ -19,43 +18,43 @@ env:
 
 jobs:
   contract:
-    name: PowerIO.jl against this capi
+    name: PowerIO.jl against this C ABI
     runs-on: ubuntu-latest
     env:
       # secrets can't be read in `if:` directly; surface it as an env var to gate on.
-      TANDEM_TOKEN: ${{ secrets.TANDEM_REPO_TOKEN }}
+      POWERIO_JL_TOKEN: ${{ secrets.POWERIO_JL_REPO_TOKEN }}
```
qian-harvard: **🟠 CI gate may be silently disabled (PLAUSIBLE — please verify).** This renames the gate secret `TANDEM_REPO_TOKEN → POWERIO_JL_REPO_TOKEN`. Every real step is gated `if: env.POWERIO_JL_TOKEN != ''`; if the secret resolves empty the job prints a skip notice and **passes green**. If the GitHub repo/org Actions secret was not renamed in lockstep, the PowerIO.jl C-ABI / Network-JSON contract test is now dead and a breaking `pio_*` change would ship untested. **Verify:** check eigenergy/powerio (and org) Actions secrets for `POWERIO_JL_REPO_TOKEN` vs a lingering `TANDEM_REPO_TOKEN`, or watch this PR's CI for the skip notice.

pyproject.toml
```diff
@@ -53,5 +57,5 @@ Repository = "https://github.com/eigenergy/powerio"
 manifest-path = "powerio-py/Cargo.toml"
 module-name = "powerio._powerio"
 python-source = "python"
-features = ["extension-module"]
+features = ["extension-module", "gridfm"]
```
qian-harvard, line 60: **🟠 Default wheel always compiles gridfm → the absent-gridfm path is no longer CI-tested.** With `features = ["extension-module", "gridfm"]`, every CI build has `_has_gridfm == True` (test, test-gridfm, validation, wheels, sdist). So `test_gridfm_absent_raises_clean_importerror` (test_powerio.py:529) hits `if HAS_GRIDFM: pytest.skip(...)` in all jobs, and the `_require_gridfm()` ImportError fallback (`__init__.py:103-114`) is exercised by none. The `test` and `test-gridfm` jobs become near-duplicates. Consider keeping one CI build with gridfm compiled out so the absent-path assertion actually runs.

README.md
```diff
@@ -1,212 +1,194 @@
...
 
-Across these cases powerio parses faster than ExaPowerIO, 62–96× PowerModels'
-parser, and ~14× pandapower's `.m` reader. It round trips byte for byte (verified at
-54 MB, 192768 buses); the other two don't. Parse, conversions, and Y_bus are checked
-value for value against all three (`benchmarks/run_validation.sh`).
+`powerio-capi` exposes parse, query, conversion, JSON transport, normalization,
+and numeric table extraction through `pio_*` functions. The public header is
+[powerio-capi/include/powerio.h](https://github.com/eigenergy/powerio/blob/main/powerio-capi/include/powerio.h).
+Build with `--features arrow` to enable `pio_export_arrow` over the
+[Arrow C Data Interface](https://arrow.apache.org/docs/format/CDataInterface.html).
```
qian-harvard, line 159: **🟠 Overstated coverage.** "The Rust test suite covers … C ABI behavior" sits directly above `cargo test`, but `Cargo.toml` `default-members` excludes `powerio-capi` (19 `#[test]`s) — `cargo test` never compiles or runs them. Add `cargo test -p powerio-capi` (or `--workspace`) to the documented commands. (The Python-bindings clause is covered by the `pytest python/tests` line already in the block.)

AGENTS.md
```diff
@@ -140,35 +140,35 @@ benchmarks/                  # parse benchmarks + Julia validation harnesses
...
-- **DC-OPF is bus-indexed.** Generation is nodal (`p_g ∈ ℝⁿ`), so `Q`, `c`, and bounds are length-n (zero at load buses), scattered from generator space through `C_g`; gen-space vectors (`OpfInstance::gen_costs`) ride along as provenance. Cost map: MATPOWER `c2 p² + c1 p` → `q = 2c2`, `c = c1`. Per-unit by default (`Units::PerUnit` scales `q` by `base²`, `c` by `base`).
+- **DC OPF Laplacian.** `L = A diag(b) Aᵀ` is built from the same `A`, `b` factors `build_incidence` returns (so `L` and the reweighted `L₁` share a factorization), and equals `build_bprime` in the XB scheme. Default `b = 1/x` (paper-pure); `DcConvention::Matpower` uses `1/(x·τ)` plus a phase-shift injection.
+- **DC OPF is bus indexed.** Generation is nodal (`p_g ∈ ℝⁿ`), so `Q`, `c`, and bounds are length n (zero at load buses), scattered from generator space through `C_g`; gen-space vectors (`OpfInstance::gen_costs`) ride along as provenance. Cost map: MATPOWER `c2 p² + c1 p` → `q = 2c2`, `c = c1`. Per-unit by default (`Units::PerUnit` scales `q` by `base²`, `c` by `base`).
```
qian-harvard, line 164: **🟠 Doc contradicts the code.** "Exactly one `BusType::Ref` is required" is the wrong general rule. `indexed.rs` exposes `reference_bus_indices()` (a *set*) and `check_reference_coverage()` (one reference *per connected component*), and `sensitivity.rs` in this PR grounds one row/col per reference. `reference_bus_index()` is only the exactly-one *convenience* query. The accurate rule was in the now-deleted `CLAUDE.md` ("Reference buses are a set … one per connected component"). As written this would lead a contributor to reject valid multi-island / distributed-slack networks.

AGENTS.md
```diff
@@ -8,24 +8,24 @@ A Cargo workspace of Rust crates plus a Python package. Parses power network
...
+  `pio_export_arrow`, an Arrow C Data Interface export.
 
-`Network` is the one canonical model (format-neutral, loads/shunts first-class);
-`IndexedNetwork` is the dense-indexed analysis view derived from it.
+`Network` is the one canonical model (format neutral, loads/shunts first class);
+`IndexedNetwork` is the dense indexed analysis view derived from it.
 
 Formats. Readers: MATPOWER `.m`, PowerModels JSON, PSS/E `.raw` (v33),
-PowerWorld `.aux`. Writers: those plus EGRET JSON. Every format meets at
+PowerWorld `.aux`. Writers: those plus egret JSON. Every format meets at
```
qian-harvard: **🟠 egret documented as write-only — contradicts the code and this PR's own README.** egret has a full reader (`parse_egret_json`, wired at `format/mod.rs:150`, byte-exact round-trip-tested in `roundtrip_formats.rs`), and README's fidelity table in this same PR lists egret as a reader. But this line omits egret from "Readers", line 92 labels `egret.rs` a "writer", `benches/parse.rs:51` calls it "write-only", and the PyO3 `parse_str` docstring (`powerio-py/src/lib.rs:574`) lists only 4 formats. The correct "all read and write" statement was in the deleted CLAUDE.md. Please fix all four spots.

powerio/benches/parse.rs
```diff
@@ -48,7 +48,7 @@ fn bench_roundtrip(c: &mut Criterion) {
 }
 
 // The readable non-MATPOWER formats, paired with the writer that produces a
-// fixture for them. EGRET JSON is write-only, so it isn't here.
+// fixture for them. egret JSON is write-only, so it isn't here.
```
qian-harvard: egret JSON is **not** write-only — `parse_egret_json` is a full reader (re-exported at `lib.rs:45`, byte-exact round-trip-tested in `roundtrip_formats.rs`). The stated rationale for excluding egret from this bench array is false; it's benchable exactly like the other three readers.

powerio-matrix/Cargo.toml
```diff
@@ -26,7 +29,7 @@ name = "powerio_matrix"
 path = "src/lib.rs"
 
 [dependencies]
-powerio = { path = "../powerio" }
+powerio = { path = "../powerio", version = "0.1.0" }
```
qian-harvard: **🟠 Release version isn't single-sourced.** This newly-added `version = "0.1.0"` path-dep pin is one of **9 sites** holding the version (5 crate `[package].version` + 4 path-dep pins), alongside duplicated `edition`/`rust-version`/`license`/`repository`/`homepage`. A bump must touch all in lockstep; miss a caret pin and `cargo publish` rejects the dep or ships a wrong version — exactly the drift a release-prep PR should remove. Add `[workspace.package]` + `[workspace.dependencies]` in the root and inherit via `version.workspace = true`. (Also: `homepage` is byte-identical to `repository` in 3 crates.)

powerio-capi/include/powerio.h
```diff
@@ -1,20 +1,23 @@
-/* powerio C ABI — parse any power system case format, query it, convert
- * losslessly, and extract the numeric tables for matrix assembly.
+/* powerio C ABI: parse any power system case format, query it, convert
```
qian-harvard, line 1: **Hygiene — generated header drifts from its source-of-truth.** This "checked in and generated" preamble line-wraps differently from the `cbindgen.toml` `header` block (e.g. `convert it,` vs `convert` / `* it,`), so re-running the documented `cbindgen … --output include/powerio.h` would rewrite it. The PostToolUse hook that auto-regenerated this header was deleted with `.claude/settings.json`, and CI only checks `pio_*` *symbol* parity (rust.yml), not header content — so the drift is unguarded.

.github/workflows/validation.yml
```diff
@@ -2,7 +2,7 @@ name: Validation
 
 # Cross-tool correctness gate: run benchmarks/run_validation.sh (the same pass/fail
 # matrix the maintainer runs locally) and fail the PR on any regression. Only the
-# machine-independent correctness checks run here — the speed benchmarks are
+# machine independent correctness checks run here. The speed benchmarks are
```
qian-harvard, line 5: **Hygiene — dangling reference.** This comment points to "the refresh-benchmarks skill", but this PR deletes `.claude/skills/refresh-benchmarks/SKILL.md` (and `.claude/` is now gitignored). Combined with the removal of README's `BENCH:speed-main` markers, the benchmark-refresh tooling (`render_tables.py`, still on disk) is now both undocumented and broken — see the summary comment. Reword or drop the parenthetical.

qian-harvard (commented):
## Automated code review — release-prep PR #69

_Recall-mode review (49 agents: 9 finder angles → adversarial verify → gap sweep). 15 distinct findings; the load-bearing ones were re-verified against the code. Inline comments are attached; two findings that can't be anchored inline are below._

### 🔴 Should block the release
1. **gridfm export rejects legitimate `±Inf` limit fields** (`gridfm.rs:337-347`) — see inline. Breaks the core GridFM export path for PowerModels/PGLib cases with unbounded limits that exported fine before this PR.
2. **Documented benchmark-refresh tooling is broken** (`benchmarks/render_tables.py:104,118` — file untouched by the PR, so no inline anchor). The PR removed README's `<!-- BENCH:speed-main -->` markers, but `render_tables.py` still maps `speed-main → README.md` (:118) and `splice()` does `raise SystemExit("…marker BENCH:speed-main not found; refusing to write")` (:104). With `speed_julia.json` present, `python benchmarks/render_tables.py` (and `--check`) aborts before writing anything — **reproduced**. Restore the README marker block, or update `render_tables.py` to drop the README/speed-main target. (Its how-to skill was also deleted — see the `validation.yml` inline.)

### 🟠 Validation-surface / metadata (in scope for this PR)
3. PowerIO.jl C-ABI contract test may now silently skip — secret rename, **PLAUSIBLE, please verify** (inline on `julia-binding.yml`).
4. Default wheel always compiles gridfm → absent-gridfm path no longer CI-tested (inline on `pyproject.toml`).
5. README overstates `cargo test` coverage (C ABI tests excluded by `default-members`) (inline on `README.md`).
6. `AGENTS.md:165` reference-bus rule contradicts the code (inline).
7. egret documented as write-only in 4 surviving spots (inline on `AGENTS.md`, `parse.rs`).
8. Release version duplicated across 9 manifest sites — no `[workspace.package]` (inline on `powerio-matrix/Cargo.toml`).

### 🟡 Hygiene / cleanup
9. `powerio.h` preamble drifts from `cbindgen.toml` (inline). 10. Dangling `refresh-benchmarks` reference (inline on `validation.yml`). 11. `_powerio.pyi` declares gridfm fns unconditionally (inline). 12–13. gridfm `base_mva` / gencost re-implementations (inline). 
14. **Unreferenced assets:** `docs/assets/powerio-hero.svg` and `powerio-logo.svg` are added but nothing references them (README uses only the PNGs) — dead weight that can drift from the rendered PNGs.

python/powerio/_powerio.pyi
```diff
@@ -69,11 +75,32 @@ class PyCase:
     def write_dcopf_bundle(
         self, out_dir: str, convention: Optional[str] = ..., units: Optional[str] = ...
     ) -> dict: ...
+    def write_gridfm(
```
qian-harvard, line 80: **Cleanup.** `write_gridfm` (and module-level `write_gridfm_batch` at line 91) are declared unconditionally, but both are `#[cfg(feature = "gridfm")]`-gated in `lib.rs`. On a documented non-gridfm build, a type checker/IDE reports them as valid while direct `_powerio` access `AttributeError`s at runtime. A `.pyi` can't express a cfg-conditional surface — add a comment noting these require the gridfm feature.

powerio-matrix/src/io/gridfm.rs
samtalki: Fixed in afe4754: limit fields (gen p/q limits, branch angmin/angmax/rate_a, bus vmax/vmin) now reject only NaN; ±Inf reaches Parquet. Regression tests added (unbounded_limits_export_as_infinity, nan_limit_still_errors). Kept validation over out-of-service generators: their values land in the columns either way, so NaN there would still poison the dataset.

powerio-matrix/src/io/gridfm.rs
samtalki: Fixed in afe4754: Network::check_base_mva (powerio/src/network.rs) is now the single check; to_normalized and the gridfm export both call it.

powerio-matrix/src/io/gridfm.rs
samtalki: Fixed in afe4754: the loop now validates the exact (cp0, cp1, cp2) tuple gridfm_cost returns, so the error names the destination column and non-representable rows are covered by the zeros they actually export. The unwrap/re-wrap is gone with it.

.github/workflows/julia-binding.yml
samtalki: Confirmed real: the secret was never set after the rename, and run 27226729984 skipped green. Fixed in afe4754 with a probe step that falls back to an unauthenticated clone, so the contract test self-enables the moment PowerIO.jl goes public; the secret remains an optional interim override.

pyproject.toml
samtalki: Fixed in afe4754: the plain test job now seds the gridfm feature out of pyproject before building, so test_gridfm_absent_raises_clean_importerror and the _require_gridfm fallback run on 3.9 and 3.12; test-gridfm and the wheel jobs keep the full released build. The jobs are no longer near-duplicates.

README.md
samtalki: Fixed in afe4754: the Validation section now states the C ABI crate carries its own tests outside the default members, and 'cargo test -p powerio-capi' is in the documented command block (also in AGENTS.md).

AGENTS.md
samtalki: Fixed in afe4754: restored the set-semantics rule (one reference per connected component, reference_bus_index as the exactly-one convenience query) as its own bullet.

AGENTS.md
samtalki: Fixed in afe4754: AGENTS.md now says all five formats read and write, and the layout comment labels egret.rs reader + writer.

powerio/benches/parse.rs
samtalki: Fixed in afe4754: comment corrected and egret-json added to the bench FORMATS array, so the reader is benchmarked like the other three.

powerio-matrix/Cargo.toml
samtalki: Fixed in afe4754: [workspace.package] + [workspace.dependencies] in the root now single-source version/edition/rust-version/license/repository/homepage and the intra-workspace version pins; the five crates inherit. A release bump is one line.

powerio-capi/include/powerio.h
samtalki: Could not reproduce against the current cbindgen.toml: regenerating with cbindgen 0.29 produces no preamble diff (verified while regenerating for the Arrow docstring change in afe4754). Looks stale against an earlier revision. Agreed the deleted hook leaves content drift unguarded; the README documents the regen command and CI keeps symbol parity.

.github/workflows/validation.yml
samtalki: Fixed in afe4754: the comment now points at benchmarks/render_tables.py + RESULTS.md instead of the deleted skill. The render_tables speed-main blocker from the summary was already resolved in 6d6003b (script retargeted to RESULTS.md only; --check passes).

python/powerio/_powerio.pyi
samtalki: Fixed in afe4754: both write_gridfm and write_gridfm_batch stubs now carry a comment that they exist only when the extension is compiled with the gridfm feature (the released wheel is) and raise AttributeError otherwise. A .pyi cannot express the conditional surface itself.

samtalki:
Re the two summary-only findings: render_tables.py was already fixed in 6d6003b (speed-main target removed, README markers dropped intentionally; python benchmarks/render_tables.py --check passes). The hero/logo SVGs are kept deliberately: they are the editable source for the rendered PNGs the README embeds. Everything else is addressed in afe4754 (see inline replies); the same commit also adds the release-binaries workflow, a tokenless fallback for the PowerIO.jl contract test, and a MATPOWER preallocation cap.

powerio-capi/include/powerio.h
samtalki: please open an issue if you continue to see this after I merge @qian-harvard

PR #77, opened 2026-06-10
refactor(api): canonicalize parse/convert names across Rust, Python, C (#75)

### LINKED ISSUE #75: More canonicalization
Urgent follow up to be completed either **after** PR #71 merges, or on that branch before it merges.

Several areas of canonicalization still need to be ironed out.
 
- Worst one: Python parse_matpower(path) takes a path while Rust parse_matpower(text) takes text. Recommend deleting Python parse_matpower/parse_matpower_string (parse_file/parse_str cover both).
- Rust has parse, parse_file, and read_path for one operation. Recommend: delete parse, fold the from override into parse_file (matching every binding's signature), delete read_path.
- Python aliases parse/convert/write are labeled compatibility with nothing to be compatible with. Recommend deleting all three, keep Case = Network.
- C handle is PioCase while the taxonomy says Network everywhere. 
- source_format strings ("PowerModelsJson", "EgretJson") don't round-trip through target_format_from_name, so net.to_format(other.source_format) fails for exactly those two. Recommend accepting powermodelsjson/egretjson as alias tokens.
- Python element drift: dict key type (Rust/JSON/Julia say kind), property gens (everywhere else generators), property n (vs n_buses). Recommend aligning all three.

Relates to #74

cameronkhanpour:
Closes #75. A pre-release API canonicalization pass aligning every binding with the [docs/languages.md](docs/languages.md) taxonomy. **Clean breaking changes** (no deprecation shims), **no behavior changes** — pure renames/deletions.

### Issue #75 checklist

- [x] **Python `parse_matpower`/`parse_matpower_string` deleted** — they took a *path*/*text* inconsistently with Rust; `parse_file`/`parse_str` cover both.
- [x] **Rust `parse`/`read_path` deleted; `from` folded into `parse_file(path, from)`** — one path parser matching every binding's signature. All call sites migrated.
- [x] **Python `parse`/`convert`/`write` aliases deleted** — including the `Case.write()` method; `Case = Network` kept.
- [x] **C handle fully renamed to "network"** — `PioCase` → `PioNetwork`, `pio_case_free` → `pio_network_free`, and the `case` handle parameter → `net` (updated to a full rename per @samtalki's review; PowerIO.jl is pre-release too, so its `ccall` is updated separately).
- [x] **`source_format` strings round-trip** — `target_format_from_name` now accepts `powermodelsjson`/`egretjson`, so `net.to_format(other.source_format)` works for PowerModels/egret (the only pair that failed).
- [x] **Python element drift aligned** — bus dict key `type` → `kind`, property `gens` → `generators`, `n` → `n_buses`.

### Notes
- `DenseNetwork.n`/`.m`/`.ng` (the dense matrix-dimension triple) are left as-is — distinct from the `Network.n_buses` property.
- C header regenerated with cbindgen 0.29.4; header parity verified.

### Verification (all run locally)
- **Rust:** `cargo build` (powerio/matrix/cli/capi incl. `--features arrow`), `cargo test -p powerio -p powerio-matrix -p powerio-capi` (all pass, incl. the new `source_format_strings_round_trip_to_a_target`), `clippy --all-targets` clean, `fmt --check` clean.
- **C ABI:** cbindgen header parity verified; **`smoke.c` compiled and run end-to-end → `C ABI smoke test OK`** with `PioNetwork`/`pio_network_free`.
- **Python:** `maturin develop` + **`pytest python/tests` → 68 passed, 2 skipped** (skips need `torch`/the gridfm-absent path). Covers the deleted aliases, `parse_file`/`parse_str`, `n_buses`/`generators`/`kind`, the MCP server, and the new `test_source_format_round_trips_through_to_format`.


samtalki (changes requested):
The C API rename does not need to be type only, and we shouldn't necessarily be content with the docs as the source of truth. The point here is to get the names 99% right the first time before release. Note that PowerIO.jl is not a separate repo in principle and we do not need to worry about breaking changes on that front yet because we are still pre-release there, too.

cameronkhanpour:
@samtalki done — switched to a **full** C rename in 9f1859b, not type-only:

- `PioCase` → `PioNetwork` (handle type)
- `pio_case_free` → `pio_network_free` (the free function)
- the `case` handle parameter → `net` (+ the internal `make_case`/`finish_case`/`case_ref` helpers and the handle doc prose)

Kept "case" only where it's the domain term (a power-system *case format*/*file*), not the handle. PowerIO.jl's `ccall((:pio_case_free, ...))` will need the matching `:pio_network_free` — agreed it's fine to break pre-release; I updated the raw-ccall example in the capi README too.

Header regenerated with cbindgen (parity verified) and the C smoke test passes against the renamed symbols.

PR #106, opened 2026-06-11
issue-47

### LINKED ISSUE #47: Full native compatibility with pandapower and pypsa
Title. Include in the benchmarks if possible.

qian-harvard:
added pandapower and PyPSA converters


benchmarks/RESULTS.md
```diff
@@ -122,14 +122,24 @@ What each column checks:
   the per-bus comparison. `_m2ppc` is used instead of `from_mpc` because it runs
   before the `from_ppc` step that raises on dclines and parallel branches.
 
+The issue-47 converter suite adds two more oracle legs to
```
samtalki: Lines 125-134 read a little bit like a changelog, can we condense?

powerio/src/format/pandapower.rs
samtalki, line 1: pandaspower JSON is a great choice; however, I'm wondering if we might also want to include other pandapower format expressions, like excel, pickle, or SQL?

https://pandapower.readthedocs.io/en/latest/file_io.html#pickle

Up to you if you think this is worthwhile right now @qian-harvard

powerio/src/format/pypsa.rs
samtalki, line 1: Same comment as panda power, but this is maybe even more important for PyPSA:
PyPSA supports NetCDF and HDF5, as well as csvs, which is quite unique to PyPSA. 

What types get used most across pandapower and PyPSA? Are JSON and CSV, respectively, the "canonical" best choice for the two?

powerio/tests/convert.rs
samtalki, line 1: general comment: make sure to use `rustfmt`

python/powerio/__init__.py
```diff
@@ -1,15 +1,18 @@
 """powerio: lossless power system case file IO, conversion, and matrices.
 
-Parse MATPOWER, PSS/E, PowerWorld, PowerModels JSON, and egret JSON into one
-format-neutral case; write it back byte exact; convert between formats; and
-pull the sparse matrices and graph views solvers need::
+Parse MATPOWER, PSS/E, PowerWorld, PowerModels JSON, egret JSON, pandapower
+JSON, and PyPSA CSV folders into one format-neutral case; write retained text
```
samtalki, line 4: related to previous comments. Are these the canonical formats? I'm worried about PyPSA in particular because I thought netCDF was the main format for pypsa

samtalki (changes requested):
Looking good. One question: Are we happy with just these two types (json,csv) or do we want this PR to be a slightly longer-term effort to be merged after the powerworld feature? up to you.

benchmarks/RESULTS.md
qian-harvard: Let's do it after the powerworld feature

powerio/src/format/pypsa.rs
qian-harvard: we can start with JSON and CSV for current version. But for large scale grid, for sure we need HDF5

benchmarks/RESULTS.md
samtalki: I think it would be awesome to release them at once honestly... but it would add a lot of new tests we would have to do

powerio/src/format/pypsa.rs
samtalki: @claude open an issue for HDF5 and NetCDF

benchmarks/RESULTS.md
qian-harvard: Condensed in 68abdc5 — the pp-json and pypsa legs are now two entries in the existing oracle list, present tense, no issue reference.

powerio/src/format/pypsa.rs
qian-harvard: Opened #107 for NetCDF/HDF5 support, and added a scope note in docs/python.md and the package docstring (68abdc5): CSV folders are PyPSA's native static component format and cover what powerio models; time-series NetCDF/HDF5 is tracked in #107.

powerio/tests/convert.rs
qian-harvard: `cargo fmt --check` passes clean across the workspace (verified on 68abdc5).

python/powerio/__init__.py
qian-harvard: Scope note added in 68abdc5 — pandapower JSON is pandapower's canonical round-trip format (`to_json`/`from_json`), and PyPSA CSV is its native static component format; NetCDF/HDF5 deferred to #107.

benchmarks/RESULTS.md
samtalki: @qian-harvard thanks, resolving conflicts and doing a review and then I'll get ready to merge

samtalki:
@qian-harvard reviewing now

samtalki (commented):
Structured review of this PR: five specialized review agents (code, tests, comments, silent failures, type design) over the full diff, with every finding adversarially verified against the code before reporting. 59 raw findings, 57 confirmed, 2 refuted. Condensed here.

## Critical (8)

1. `powerio/src/format/pypsa.rs:192,431` PyPSA transformer `r/x` are per unit on the transformer's own `s_nom`. The reader copies them onto the system base unconverted (and `base_mva` defaults to 1.0 for native folders); the writer mirrors the error and emits `s_nom = rate_a`, which is 0.0 for unlimited MATPOWER transformers. Both directions need the `s_nom` rebase. The benchmark filters transformers out of its line check, so nothing catches this.
2. `pypsa.rs:468` Storage write derives `p_nom = max(charge, discharge, energy_rating)`, mixing MW with MWh. A 25 MW / 4 h unit round trips to 100 MW / 1 h: ratings inflated 4x, duration collapsed.
3. `pypsa.rs:318` `buses.csv` rows are keyed by bus name while every element table references the raw numeric id, so PyPSA cannot import the output when any bus is named. The pandapower reader in this PR always populates names, so pandapower to PyPSA produces unimportable folders.
4. `pypsa.rs:54,590` Bus id synthesis (numeric name, else position) collides on mixed names: a valid folder with buses "2" and "x" fails with "duplicate bus id 2" (reproduced). The numeric fallback in `bus_name_ref` silently attaches a reference to a nonexistent bus onto whichever bus got that synthetic id.
5. `powerio/src/format/pandapower.rs:762` Every bus reference defaults to 0 via `usize_or(key, 0)`. Missing, null, or float encoded cells (the frame's `dtype` map is ignored; pandas upcasts integer columns routinely) silently rewire elements to real bus 0 and the parse succeeds with wrong topology.
6. `pandapower.rs:59` The 0-based pandas index maps straight to `BusId`, so every native pandapower file yields `BusId(0)`; converting onward emits bus number 0, which MATPOWER and PSS/E reject (reproduced). Needs an explicit index shift on read and write. The PyPSA reader enforces id positivity; this one does not.
7. `pandapower.rs:111`, `pypsa.rs:28` Whole tables are dropped silently on read: `sgen` (how most real pandapower nets carry distributed generation), `switch` (open switches become in service lines), `trafo3w`, `storage`, `dcline`; PyPSA `links.csv` (HVDC). Root cause: the read path returns a bare `Result<Network>` with no warnings channel, while the write path has one.
8. `pandapower.rs:199` Electrically significant columns silently ignored: `parallel` (parallel=2 halves the parsed admittance vs pandapower), `tap_neutral` (tap formula omits it), ZIP load composition, `g_us_per_km`, trafo magnetizing branch, `tap_side`. Valid input parses into a different electrical model.

## Important (20, condensed)

**Conversion.** Storage writer drops `p_set`/`q_set`/`state_of_charge_initial` that the reader consumes (`pypsa.rs:463`). Generators with `control="PQ"` (PyPSA's default) silently retype their bus to PV (`pypsa.rs:113`). Piecewise and cubic or higher costs are dropped with a bare `continue` and no warning, contradicting the fidelity doc (`pandapower.rs:581`).

**Robustness.** `Row::index_usize` panics on a frame with rows but a missing or short index, reachable from `parse_str`; in Python this surfaces as a PanicException instead of PowerIOError (`pandapower.rs:670`). `read_frame` coerces malformed columns and rows to defaults and never checks `is_multicolumn` (`pandapower.rs:715`). PyPSA folders bypass the zero bus rejection every other reader gets (`format/mod.rs:163`).

**Surfaces.** MCP `_parse` drops the `format` argument for path inputs the new docstrings advertise (`python/powerio/mcp/server.py:56`). The `pio_parse_str` format list in the doc comment and generated header is now false by omission (`powerio-capi/src/lib.rs:188`).

**Tests and CI.** No test reads a file pandapower or PyPSA actually produced; both oracles validate the write direction only. The `ext_grid` read branch is dead code in every test because the writer always emits an empty `ext_grid` frame, yet real nets put the slack there. The trafo read path (vk/vkr reconstruction, tap math) runs in zero tests. Round trips use case9 only: no shunts, no transformers, no out-of-service elements. Generator costs are asserted by nothing; a cp0/cp2 column swap passes every test and validator. `validation.yml`'s oracle import gate omits pypsa, so a broken install SKIPs every pypsa leg and CI stays green. `validate_pypsa.py` fails open when the line/transformer split mismatches.

**Docs.** `docs/format-fidelity.md` describes `ext_grid` writer behavior that does not exist, claims unsupported costs and extras "are reported as warnings" (false), says line charging "uses siemens" (the pandapower codec uses nF/km via a frequency dependent conversion), and leaves the reader's silent table subset undocumented. `parse_file`'s docs predate the three-way JSON sniff and directory inputs. New prose has hyphenated open compounds, including "power-flow".

## Division of work

**I am taking everything above.** Fixes for all critical and important items will land on this branch shortly, including the read side warnings channel, the bus identity reworks for both formats, genuine native file fixtures, and the CI gate fixes. Three suggestion level findings dissolve into that pass (`server.py:222` docstring vs `_parse`, the `format-fidelity.md:104` hyphen, and the `pandapower.rs:286` ext_grid writer contract).

@qian-harvard the remaining suggestion level findings are yours. Complete list, nothing omitted:

1. `pandapower.rs:543` `f_hz` is dropped on the non echo write path: a 60 Hz net is rewritten as a 50 Hz file with compensated `c_nf_per_km`. The Y bus is preserved exactly, but the silent relabeling deserves a warning or a doc note.
2. `pandapower.rs:36` No negative tests for malformed input in either reader: wrong `_class`, missing `_object`, frame whose `_object` is not valid JSON, missing `buses.csv`, dangling references.
3. `format/mod.rs:261` `parse_file` PyPSA routing untested: the alias normalization and the `network.csv` directory auto detect never execute in any test.
4. `powerio-capi/src/lib.rs:440` `pio_write_pypsa_csv_folder` has no execution coverage; extend `examples/smoke.c` (CI compiles and runs it).
5. `pandapower.rs:174` The `max_i_ka` sentinel (>= 99999 kA reads as unlimited) and the inverse zero rating guard are asserted by nothing.
6. `pypsa.rs:463` Once the storage writer is fixed, pin the storage write/read contract with a test: which fields round trip, which warn.
7. `python/powerio/mcp/server.py:111` The new format claims in the MCP docstrings have no cases in `test_mcp.py`.
8. `benchmarks/RESULTS.md:124` documents pp-json and pypsa columns that appear in no published results matrix; publish a run or cut the bullets.
9. `pypsa.rs:533` `read_csv_optional` treats any stat error as absence: a permission denied `generators.csv` reads as a network with zero generators.
10. `pypsa.rs:28` Decide whether PyPSA `Store` components map to `Network.storage`; after my pass the reader will warn on unread tables, but the mapping decision is open.
11. `pypsa.rs:537` The CSV writer quotes embedded newlines that the reader's `lines()` based parser cannot read back.
12. `pypsa.rs:504` Two parallel row table micro codecs (`DataFrame`/`Row` vs `CsvTable`/`CsvRow`) plus a copy pasted helper trio; consolidate. `CsvTable` adds nothing over `Vec<CsvRow>` and clones every header string per row.
13. `powerio-capi/src/lib.rs:440` returns 1 on success / 0 on error, a third return convention in the ABI alongside pointer/NULL and 0/-1; align it before bindings guess wrong.
14. `format/mod.rs:261` `is_pypsa_csv_name` is a second ad hoc alias matcher outside `target_format_from_name`, whose doc still claims to be the one place bindings share.
15. `pandapower.rs:736` `poly_cost` map keyed by `(String, usize)` with a fresh `String` allocated per lookup; the `et` column is a small closed domain that wants an enum, and non gen costs are read then dropped.
16. `pandapower.rs:254` "writer v1" labels in user facing warnings and docs will rot; no other writer in the repo is versioned this way.
17. `docs/format-fidelity.md:32` The validation overview says every classic reader and writer runs under an oracle but omits that the pandapower JSON reader has none.
18. `docs/python.md:64` The CSV folder component list omits transformers and storage units, both of which the writer emits and the reader consumes.
19. `README.md:227` Two consecutive sentences open with "The benchmark validation suite"; merge them.
20. `format/mod.rs:130` `target_format_from_name`'s doc explains the `powermodelsjson`/`egretjson` aliases but omits the new `pandapowerjson` alias that exists for the same round trip reason.

Two findings were raised and refuted during verification (short CSV row padding, `write_pandapower_json` source echo duplication); they are not issues.

qian-harvard:
Suggestion-level findings from the structured review are handled in 7eb3ee7. Item by item:

**Done (17 of 20):** 1 (f_hz: documented in format-fidelity.md + a code comment — `Network` carries no frequency, so a conditional warning isn't possible), 2, 3, 4 (smoke.c now writes a PyPSA folder and checks the NULL-handle failure), 5, 7, 8 (bullets cut from the published-matrix column list; the legs are noted as not yet published — see below), 9 (only `NotFound` reads as an absent table now), 11 (the reader is a full-record CSV parser that honors quoted newlines, with a round-trip test), 13 (`pio_write_pypsa_csv_folder` now returns 0/-1; no external callers existed), 14, 15 (`CostElement` enum; non-gen/ext_grid `et` rows skipped explicitly), 16, 17, 18, 19, 20.

**Deferred to your pass (3 of 20):** 6 and 10 depend on the storage-writer fix and the read-side warnings channel you're landing; 12 (codec consolidation) rewrites the same code your critical fixes touch, so doing it now would just conflict. Happy to pick all three up after your push.

**One drive-by fix + a data point for your critical pass:** the pandapower JSON writer was missing the `tap_step_degree` trafo column, which `makeYbus` requires once a trafo table exists — every transformer fixture failed to import. Fixed in 7eb3ee7. With that fixed, I ran `validate_pandapower_converter.py` over the full fixture list (pandapower 3.1.2 locally, not the pinned 3.2.2): case9/case30/t_case9_dcline/t_case9_oos/case5_pjm match, but the transformer cases still mismatch on Y_bus — case118 (30 entries, worst |Δ|=3.9), pglib case14_ieee (8 entries, worst |Δ|=0.6), case2869pegase (1042 entries, worst |Δ|=298), and case57 hits a FloatingPointError inside pandapower. Consistent with your critical finding 8 (trafo tap/magnetizing model), so leaving those to your pass.

qian-harvard:
CI status after 17d529f and 241fd01:

- **Rust / Docs**: fixed — clippy lints in the new tests and the rustdoc private-intra-doc-link.
- **Validation, pypsa leg**: green on all ten fixtures (verified locally under PyPSA 1.2.2). Root cause was critical finding 3 from the structured review: buses.csv is keyed by name while element tables wrote raw numeric ids, so any case with named buses (every IEEE fixture via the MATPOWER bus_name block) was unimportable. Element tables now write the buses.csv key. The validator also mirrors the writer's 0 kV zbase guard instead of dividing by zero.
- **Validation, pp-json leg**: half fixed. The pinned pandapower 3.2.2 renamed the load ZIP columns (`const_z_percent` → `const_z_p_percent` et al.), and the trafo frame was missing `tap_step_degree` — the writer now emits both ZIP generations and the tap column, and all six non-transformer fixtures pass under 3.2.2.

**Still failing (5 pp-json cells), for the claimed critical pass:** the transformer fixtures. Two distinct causes, diagnosed locally:
1. case14/case57: those cases carry `base_kv = 0` on every bus; the writer emits `vn_kv = 0`, and pandapower's ohm-based Y_bus divides by `vn²`. Making these round-trip needs a synthesized nominal voltage written consistently into `vn_kv` and the ohm conversion (writer's pu→ohm zbase must equal pandapower's `vn²/sn` reconstruction).
2. case118/pglib_case14/case2869pegase: real Y_bus deltas on transformer entries (case118: 30 entries, worst |Δ|=3.9) — the trafo charging/magnetizing and tap model gaps from critical finding 8 (MATPOWER transformer `b` is currently dropped; `tap_side`/`tap_neutral` reconstruction).

Both are inside the trafo rework already claimed in the review's division of work, so I've left them rather than collide with that pass.

samtalki:
The critical and important fixes have landed (9db9a53 + merge 01c8082).

**What changed**

- **Read fidelity channel**: `parse_file`/`parse_str` now return `Parsed { network, warnings }`; read warnings flow through the C ABI (`pio_parse_warnings`, header regenerated), Python (`read_warnings`), the CLI (stderr), and MCP. PyPSA folders hit the same zero bus rejection as every other reader.
- **pandapower**: bus ids shift between the 0-based pandas index and `BusId`; bus references are strict and dtype aware (missing/null/negative/fractional cells error instead of defaulting to bus 0); `sgen`/`storage`/`dcline` are read (PQ generators, `Storage`, `Hvdc`); `parallel` and `tap_neutral` enter the branch model; ZIP loads, `g_us_per_km`, magnetizing branch, and lv side taps warn; `ext_grid` rows are written for Ref buses without a generator; `read_frame` validates structure; the missing index panic is a parse error.
- **PyPSA**: transformer `r/x` rebased between the `s_nom` and system bases in both directions (`s_nom = base_mva` when `rate_a` is unlimited); storage `p_nom` no longer maxes MW against MWh and the writer emits `p_set`/`q_set`/`state_of_charge_initial`; one bus key joins `buses.csv` and every referencing table so PyPSA imports named bus output; reader ids never collide and unknown references error with the offending name; `control="PQ"` leaves the bus kind alone; `links.csv` reads into `Hvdc`.
- **Tests/CI**: genuine vendored fixtures (pandapower 3.2.2 JSON, PyPSA 1.2.2 CSV folder) with value level read tests; round trips extended to case14 and t_case9_oos; gen cost round trip assertions; `validation.yml` import gate includes pypsa; `validate_pypsa.py` fails closed on a line/transformer split mismatch and checks transformer parameters.
- **Docs**: format-fidelity.md claims now match the code; `parse_file`/`pio_parse_str` docs updated.

@qian-harvard your four commits are merged in, not overwritten: where we both fixed the same thing (the buses.csv key join) the merge keeps one implementation with your extra coverage ported; your `CostElement` typing, ZIP dual column emission, alias consolidation, capi convention fix, negative tests, and doc fixes are all in, adapted to the `Parsed` API. Heads up that `parse_file` now returns `Parsed` — new code wants `.network`.

Gate on the merged tree: `cargo fmt` clean, `clippy --workspace --all-targets` zero warnings, 294 tests passing (`--exclude powerio-py`, whose cargo test link step is environmental and excluded in CI too), cbindgen header parity holds.

docs/python.md
```diff
@@ -42,6 +44,29 @@ graph = net.to_networkx()    # needs powerio[graph]
 `parse_file(path, from_=None)` reads any format (inferred from the extension, or
 forced with `from_`); `parse_str(text, format)` reads in-memory text.
 
+## PyPSA folders
+
+PyPSA CSV folders are multi-file datasets, so they use explicit read/write
```
samtalki, line 49: Same comment as above regarding Parsed and Conversion.

powerio-capi/include/powerio.h
samtalki, line 1: we should be careful about changing the C ABI

tests/data/pandapower/example.json
samtalki, line 1: edit: taking care of the license file

tests/data/pypsa/README.md
samtalki, line 1: Same as the README for pandapower. License would be nice.

docs/python.md
samtalki: all good, I get it now

samtalki (commented):
@qian-harvard this is looking good. I have some minor non-blocking comments.

samtalki (approved):
LGTM

Issue #64, opened 2026-06-08
Activate tandem CI + docs deploy: TANDEM_REPO_TOKEN secret + enable GitHub Pages


samtalki:
#54 landed two cross-repo automations. Status:

**1. Tandem contract tests.** `.github/workflows/julia-binding.yml` (here) and `CI.yml` (in PowerIO.jl) build the other repo's C ABI / binding and run the cross-repo suite, so a change to the `pio_*` surface or the `Network` JSON schema that breaks the Julia binding fails at the source. The probe step checks out `eigenergy/PowerIO.jl` anonymously when it is public; while it is private, the job needs a token with read access to it as the `POWERIO_JL_REPO_TOKEN` secret, otherwise it prints a skip notice and passes, leaving the contract unenforced.

- Action: add `POWERIO_JL_REPO_TOKEN` (fine-grained PAT or org token with read access to `eigenergy/PowerIO.jl`), or make PowerIO.jl public, whichever comes first. Mirror on the PowerIO.jl side for its checkout of this repo.

**2. Docs deploy: done.** GitHub Pages is enabled (Source = GitHub Actions); `docs.yml` builds `cargo doc --all-features` and deploys on pushes to `main`.

Neither blocks development; the version-checked C ABI handshake (`pio_abi_version`) already catches a stale or mismatched library at load time.

PR #112, opened 2026-06-12
v0.3.0 candidate: memory safety hardening + C ABI v4

samtalki:
Merges two malformed-input fixes, generalizes both bug classes, and revises the C ABI to v4. The header preamble (`powerio-capi/include/powerio.h`) is the normative statement of the v4 conventions.

## Memory safety

- **MATPOWER gencost**: a crafted NCOST (e.g. `1e20`) overflowed the row-width arithmetic and panicked on every build profile. The arithmetic now saturates and the row is rejected as a parse error. (`claude/keen-feynman-vv3049`)
- **Message truncation**: clipping a message at a raw byte count can split a multibyte UTF-8 character, leaving invalid text in the caller's buffer. Truncation now lands on a character boundary. (`claude/amazing-edison-4bjitk`)
- **`.pwd` reader**: the byte-read helpers indexed the buffer directly and relied on per-call-site bounds checks. They now return `Option`, so an out-of-range offset from a corrupt file rejects the record instead of panicking; the record scan also retains decoded coordinates rather than re-reading them. The differential oracle tests (decoded coordinates checked against same-vintage `.aux` files across the save corpus) pass unchanged.
- **Fuzz harnesses** (`fuzz/`, workspace-excluded; nightly + cargo-fuzz): matpower, psse, and powerio-json via `parse_str`; the `.pwb`/`.pwd` decoders on raw bytes. Invariant: any input yields `Ok` or a structured `Err`, never a panic. All five targets pass seeded smoke runs.
- Audited, no change needed: `.pwb` cursor reads are bounds-checked; the psse/egret/pandapower numeric casts never feed indexing; every entry point already catches panics at the boundary.

## C ABI v4 (`PIO_ABI_VERSION` 3 → 4)

v3 used three different verbs for serialization, named a handle-returning transform like the string serializers, and let extractors write past a miscounted buffer. All known consumers are this repo and PowerIO.jl, and `pio_abi_version` rejects mismatched libraries at load, so the break is cheap now. The conventions are designed so it is the last one.

| v3 | v4 |
|---|---|
| `pio_to_matpower`, `pio_to_json`, `pio_from_json` | removed; `matpower` and the new `powerio-json` snapshot are format strings into `pio_to_format` / `pio_parse_str` |
| `pio_to_normalized` | `pio_normalize` — a value transform returning a handle; `to_` re-encodes unchanged data |
| `pio_export_arrow` | `pio_to_arrow` |
| `pio_write_pypsa_csv_folder` | `pio_write_dir(net, to, dir, ...)` |
| `pio_read_gridfm`, `pio_gridfm_scenario_ids` | `pio_read_dir(dir, from, scenario, ...)`, `pio_scenario_ids(dir, from, ...)` |
| `pio_parse_warnings` | `pio_warnings` |
| `pio_reference_bus` (isize), `pio_reference_buses`, `pio_n_reference_buses` | `pio_ref_bus_index` (i64), `pio_ref_bus_indices(net, out, cap)` — a dense index, not a bus id, and named so |
| `pio_n_components`, `pio_nodal_demand`, `pio_nodal_shunt` | `pio_n_islands`, `pio_bus_demand`, `pio_bus_shunt` |
| `pio_convert_file(path, to, from)` | `pio_convert_file(path, from, to)`; new `pio_convert_str(text, from, to)` |

No format names remain in the symbol table; adding a format leaves the ABI unchanged.

Conventions:

- **Extractors (cap/count)**: each takes the caller's buffer capacity and returns the total available. Writes never exceed `cap`, so a miscounted buffer reads short instead of overflowing, and `(NULL, 0)` is a count query — the `snprintf` pattern. v3 wrote exactly `pio_n_*` elements on trust.
- **Warnings**: `pio_warnings` returns the byte length of the joined text, so a buffer can be sized exactly; v3 returned a warning count, which cannot size a buffer. Warnings attach to the handle from any constructor; only functions returning no handle (`pio_to_format`, `pio_convert_*`, `pio_write_dir`) take a `warnbuf`.
- **Errors**: caller-provided `errbuf`/`errlen` (the libpcap/curl idiom) — no library-allocated strings to free, no thread-local state.
- **Vocabulary**: a bus is a named connection point; a node is one conductor's point at a bus (the OpenDSS sense), reserved for the multiconductor surface; a branch is any two-terminal series element, lines and transformers alike. Hence `bus_demand` and `n_islands`.
- **Freeze**: existing signatures never change again. New data is new symbols; rich or multiconductor data rides the Arrow C Data Interface and the `powerio-json` snapshot, whose schemas evolve without touching a C signature.

Supporting changes: `TargetFormat::PowerioJson` makes the snapshot an ordinary format, reachable from the CLI and the converters; `write_as`/`to_format` become fallible because the snapshot rejects non-finite values rather than writing `null` (foreign JSON targets are unchanged); `powerio::write_dir` and `powerio_matrix::read_dataset_dir`/`dataset_scenario_ids` are the directory-format dispatch points; `examples/smoke.c` exercises the full v4 surface and is compiled and run in CI.

## Verification

Workspace test suite, 26 capi unit tests, header parity, the compiled C smoke binary end to end, PowerIO.jl's 180 tests against this branch's library, the PowerModels/Exa oracle matrix, and the fuzz smoke runs. Benchmarks against a main baseline show no regressions; the two reworked readers measure flat (matpower) and 1.3% faster (`.pwd`). Full table: [benchmark comment](https://github.com/eigenergy/powerio/pull/112#issuecomment-4688449098). Continuous tracking: #115.

## Numbering and pairing

No version number changes in this diff; the branch keeps its working name. Recommended release number: 0.3.0 — pre-1.0 convention puts breaking changes in the minor, and the ABI handshake remains the actual compatibility gate. Merge order: this PR, then eigenergy/PowerIO.jl#25, with binaries cut from the same commit (tandem CI inactive, #64). Follow-ups: #113 (dist surface adopts these conventions), #114 (PowerDiff field renames), #115 (benchmark tracking).


samtalki:
Lockstep Julia PR: eigenergy/PowerIO.jl#25 — merge this one first, then that, and cut binaries from the same commit.

samtalki:
Benchmark verification before merge (criterion A/B, branch vs main baseline, shared target dir, M-series local):

| benchmark | main | v4 branch | change |
|---|---|---|---|
| parse_case2869pegase (headline) | 1.665 ms | 1.662 ms | none (p=0.33) |
| parse_case57 / case118 | 28.4 µs / 75.2 µs | 28.2 µs / 75.6 µs | -0.9% / none |
| roundtrip_case2869pegase | 1.662 ms | 1.674 ms | +1.2% (see note) |
| parse_powermodels-json_case118 | 523 µs | 534 µs → 524 µs | flagged +3.5%, **not reproducible**: rerun -0.7% |
| parse_psse_case118 | 567 µs | 537 µs → 560 µs | mirror jitter, rerun: none |
| parse_powerworld / egret / aux | 388 µs / 293 µs / 2.536 ms | 386 µs / 296 µs / 2.524 ms | none |
| parse_pwb_activsg200 | 349.9 µs | 353.8 µs | +1.3% |
| **parse_pwd_activsg200** (new bench; the reworked per-byte hot loop) | 169.1 µs | 167.5 µs | **-1.3%, improved** |

Note: the only statistically significant shifts are ≤1.3% on µs-scale benchmarks whose code this PR does not touch (matpower write/roundtrip, pwb), with equal-magnitude shifts in both directions across the suite — codegen layout jitter under lto=thin/codegen-units=1, not algorithmic change. The two readers this PR actually reworked measure flat (gencost saturating arithmetic) and faster (pwd total accessors, which now also cache decoded coordinates instead of re-reading). The flagged powermodels regression vanished on an immediate rerun against the same baseline.

Continuous tracking so this stops being a manual ritual: #115.

Issue #124, opened 2026-06-15
Multiconductor to balanced lowering: design contract


frederikgeth:
v0.4.0 scope is the design contract only. The implementation is tracked in #145.

Goal: define an explicit lowering pass from `MulticonductorNetwork` to `BalancedNetwork` for tools that operate on the balanced model. This is not a merger of the two models.

Design principles:

- Keep `BalancedNetwork` and `MulticonductorNetwork` as distinct concrete IRs under the top level v1 `Network` package direction in #144.
- Start from transparent, well defined cases rather than pretending every distribution feeder has a unique positive sequence equivalent.
- Apply Kron reduction before sequence transforms when a neutral conductor is present.
- Pick and document the sequence transform convention, including whether it is power invariant.
- Reject or warn on ambiguous one wire and two wire cases.
- Preserve diagnostics explaining what was reduced, approximated, dropped, or rejected.

This issue closes when the design contract is written into the v0.4.0 release direction and #145 carries the implementation work.

samtalki:
I think this should be framed as a lowering pass rather than as a merger of the current transmission and distribution structs.

For the v0.4 API break, the direction should be:

- `powerio::Network`: the canonical static network IR container.
- `BalancedNetwork`: the current scalar positive sequence model.
- `MulticonductorNetwork`: the current `DistNetwork`, aligned with BMOPF.

Then this issue becomes `MulticonductorNetwork -> BalancedNetwork`: lower a multiconductor distribution feeder to a positive sequence or single phase equivalent for tools that operate on the balanced model. That keeps the BMOPF level canonical for distribution while still making equivalent transmission style studies possible.

frederikgeth:
I agree. It would be great to do this through PowerIO, from my perspective. Let me know if / how I can help.

frederikgeth:
Btw, OpenDSS has a method to reduce cases to positive sequence it but it’s not best practice so let’s at least be more transparent. Sequence transformation is not well defined unless there are exactly 3 voltages. For 3 phase + neutral you do Kron's reduction first. However for 2/1wire it's less obvious what the best choice is. Furthermore, there's power invariant and non power invariant transformations, and the literature isn't always clear on which one is used.

Issue #29, opened 2026-06-06
Origin, source maps, structured diagnostics, and converter tests


samtalki:
V1 compiler object dependency for #129.

The current model has `source_format` plus optional retained source text and string warnings. That is enough for current conversion, but not enough for `.pio.json` as a compiler object.

Scope:

- Replace invalid provenance states with an `Origin` style representation that distinguishes in memory objects, retained source text, folder inputs, binary inputs, and derived products.
- Add source map structures where readers can point canonical fields back to source files, lines, records, or fields.
- Replace or supplement string warnings with structured diagnostics for dropped data, defaulted data, best effort mappings, validation findings, and repair suggestions.
- Keep human readable warning text available for CLI and Python users.
- Strengthen converter tests so all pairs compare per element values where stable ids exist, not only counts and totals.

Acceptance:

- `.pio.json` can carry provenance and diagnostics without parsing free form strings.
- CLI, MCP, Python, C ABI, and Julia consumers can still show readable warnings.
- Cross format conversion tests catch moved loads, swapped limits, and lost element fields.

samtalki:
The v0.1.0 release-hardening PR closes the **writer fidelity-warning** slice of this issue: the PSS/E and PowerWorld writers now push `Conversion::warnings` when they drop fields the target can't represent — branch angle limits (`angmin`/`angmax`), generator ramp/capability data (`GenCaps`), and per-element `name` — matching the existing `dcline`/`storage`/`cost` warning pattern. The two-tier contract (same-format byte-exact; cross-format keeps maximal fidelity and warns on loss) is now honored by these writers, not just EGRET.

**Remaining here**: the `Origin` type (provenance of each field across a convert chain) and the deeper round-trip/value tests beyond the all-pairs harness.

samtalki:
This should be treated as a core dependency of the v0.4 compiler architecture.

The `Origin` and structured warning pieces are exactly what the `.pio.json` compiled object layer needs:

- source format and source text/reference without invalid state pairs
- source maps from canonical fields back to input bytes/lines/fields
- structured warnings that distinguish dropped data, defaults, best effort mappings, validation findings, and repair suggestions
- machine readable diagnostics for CLI, MCP, Python, C ABI, and Julia consumers

The v0.4 split should make `Network` the canonical static network IR container, with concrete balanced and multiconductor models underneath. `.pio.json` should package that IR with `Origin`, source maps, warnings, validation results, summaries, and optional matrix/cache metadata.

[...]

PR #143, opened 2026-06-25
[data] rich data model parity

samtalki:
## Summary
- add typed branch terminal admittance, switches, current ratings, branch solution values, HVDC cost, storage current rating, and load voltage models
- map richer PowerModels, PSS/E, pandapower, PyPSA, BMOPF, PMD, and OpenDSS fields into the typed model
- update Y bus assembly and Python, C, and Arrow exports, with tests for asymmetric charging, switches, ZIP and exponential load models, and HVDC cost scaling
- add the rich model validation tier: committed rich fixtures, the PowerModels rich oracle, and opt in local corpus reports that stay under `benchmarks/results/`

## Validation
- cargo fmt --all --check
- cargo clippy --all-targets
- cargo test -p powerio -p powerio-matrix -p powerio-dist
- cargo test -p powerio-capi
- cargo test -p powerio-capi --features arrow
- cargo check -p powerio-py
- .venv/bin/python -m pytest python/tests
- cargo test -p powerio
- bash benchmarks/run_rich_validation.sh

Related: #124, #144, #153, #29


samtalki:
@frederikgeth would appreciate your eyes on this PR as we have discussed all of this at some point:
- Added typed load voltage and nominal voltage preservation across PSS/E, PSLF, *PMD, BMOPF, and OpenDSS*
- **_Important_:** v1 `BalancedNetwork` and `MulticonductorNetwork` aliases in preparation for the introduction of the PowerIO IR in v1.0.0;
- tightens MATPOWER lossy warnings; 
- rich parity tests for exact nonzero ZIP components.
- `DistLineCode` now carries full multiconductor g_from/b_from/g_to/b_to matrices

qian-harvard:
### Code review

Found 4 issues. They share one root cause: this PR upgrades the format readers to capture rich fields (per-terminal admittance, ZIP load models) and removes their old read-side "ignored" warnings, but several writers were not updated to either emit the new data or warn that they drop it. That breaks the fidelity contract: "Whatever the target format cannot represent is reported in the `Conversion` `warnings`, never dropped silently."

https://github.com/eigenergy/powerio/blob/f83636a10030d797d67adaf4bc016552dfdfa9e2/powerio/src/format/mod.rs#L23-L26

1. pandapower JSON writer silently demotes ZIP / voltage-dependent loads to constant power. `load_frame` never reads `Load.voltage_model` and hardcodes all six ZIP columns to `0.0` with no warning, so a pandapower net with ZIP loads round-trips to pure constant power — changing power-flow results at any voltage other than 1.0 p.u. The columns are representable (the reader already builds `LoadVoltageModel::Zip` from them) and there is no pandapower write-back round-trip test.

https://github.com/eigenergy/powerio/blob/f83636a10030d797d67adaf4bc016552dfdfa9e2/powerio/src/format/pandapower.rs#L862-L877

2. PSS/E branch writer silently drops terminal conductance (GI/GJ) and the asymmetric susceptance split. The reader stores GI/BI/GJ/BJ in `charging`, but the writer emits only `num(br.b)` and hardcodes the four terminal-shunt fields to `0`, with no warning — unlike the MATPOWER writer, which warns for the same class of loss.

https://github.com/eigenergy/powerio/blob/f83636a10030d797d67adaf4bc016552dfdfa9e2/powerio/src/format/psse.rs#L289-L302

3. PyPSA CSV writer silently drops branch shunt conductance `g`. The reader maps the PyPSA `g` column into `charging.g_fr`/`g_to` (and the old warning was removed), but the `lines_csv` (L760) and `transformers_csv` (L788) headers omit `g` and no warning is emitted. PyPSA CSV supports a `g` column — the reader's own fixtures include it.

https://github.com/eigenergy/powerio/blob/f83636a10030d797d67adaf4bc016552dfdfa9e2/powerio/src/format/pypsa.rs#L758-L761

4. OpenDSS writer silently drops the ZIP `zipv` cutoff and non-ZIP load models. The 7th `zipv` element is hardcoded to `0` (losing the Vmin cutoff, e.g. 0.85), and DSS load models 3/4/6/7 read as `ConstantPower` and write back as `model=1` (the `ConstantPower` arm at L932 emits nothing), with no warning.

https://github.com/eigenergy/powerio/blob/f83636a10030d797d67adaf4bc016552dfdfa9e2/powerio-dist/src/dss/write.rs#L957-L966

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

samtalki:
Thanks, Qian. Appreciate the careful review. I pushed 0a9a63c to address these.

The four concrete gaps are fixed: pandapower writes ZIP load percentages from `Load.voltage_model`, PSS/E writes terminal charging into GI/BI/GJ/BJ, PyPSA emits branch `g` for lines and transformers, and OpenDSS preserves retained `zipv` cutoff plus retained DSS load model codes where they map to constant power.

I also added direct `Network` writer tests and warning contract coverage for rich fields that remain unrepresentable. The miss was our tests leaning too much on retained source fixtures; the new cases exercise the canonical writer paths directly.

Validation: `cargo test -p powerio pandapower`, `cargo test -p powerio psse`, `cargo test -p powerio pypsa`, `cargo test -p powerio-dist rich`, and `cargo fmt --all --check`.

frederikgeth:
the one thing that stands one is different line models and (still restrictive) generalisations may cause confusion
- Matpower has just series resistance, series reactance and shunt susceptance as inputs, which instantiates a pi model with equal shunt susceptances
- powermodels allows for a pi model with both susceptance and conductance, and the values don't need to be equal, therefore it can also capture a gamma section. Furthermore, you could deliberately also give it different susceptance/conductance values on either side, which can be practical for certain types of circuit reductions

samtalki:
Thanks Fred. I took this as a model semantics issue, not only wording.

Pushed f747a5a2. Let me know what you think about this approach.

What I went with for v0.4.0 was to make per terminal Branch::charging the canonical physical admittance when present and treat Branch::b as the MATPOWER scalar projection. MATPOWER writers/exporters now derive their scalar field from `terminal_charging().total_b()`, while PowerModels and PSS/E line records keep terminal values where the target supports them. PSS/E transformers now write total magnetizing conductance/susceptance to MAG1/MAG2 and warn when side assignment has to collapse.

I also added value tests with stale b = 0 and nonzero terminal charging for MATPOWER, egret, PowerWorld, PSLF, GridFM, C ABI, Arrow, and Python surfaces, plus docs clarifying the distinction. 

I left the public schema intact for this PR and recorded the v1 cleanup direction on #144 

I don't really love the approach I took, and definitely want the v1.0.0 stable BalancedNetwork to not expose two mutable sources of truth. We could make a serious break now at v0.4.0 if you think it's worth it or save it for v1.0.0 after the task force meeting. I won't merge without your guidance

PR #167, opened 2026-06-30
[codex] fix BMOPF generator export

samtalki:
## Summary

Fix the BMOPF writer so fixed OpenDSS generators remain in the BMOPF `generator` table instead of being converted into negative loads.

## Details

The old BMOPF writer special cased fixed P/Q generators with no cost and emitted them as negative loads. That made generators disappear from the `.dss -> bmopf` generator table. The writer now routes all distribution generators through the existing BMOPF generator path, which represents fixed dispatch with pinned `p_min`/`p_max` and `q_min`/`q_max` bounds and emits zero cost when the source has no cost.

A regression test covers the minimal DSS generator case and asserts `generator.g1` exists, `load.g1` does not, and the pinned bounds are preserved.

## Validation

- `cargo test -p powerio-dist --test bmopf`
- `cargo test -p powerio-dist`
- `cargo test -p powerio-capi --features arrow,gridfm,dist,pkg`
- `cargo fmt --all --check`

### DIFF
--- a/powerio-dist/src/bmopf/write.rs
+++ b/powerio-dist/src/bmopf/write.rs
@@ -12,8 +12,8 @@ use serde_json::{Map, Value, json};
 
 use crate::convert::Conversion;
 use crate::model::{
-    Configuration, DistGenerator, DistLoadVoltageModel, DistNetwork, DistSourceFormat,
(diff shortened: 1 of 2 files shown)


frederikgeth:
Cool, thanks!

PR #174, opened 2026-07-02
Declare the .pio.json payload schema and resolve operating point updates by identity

### LINKED ISSUE #173: Declare a v1 payload schema and identity based operating point addressing
The envelope is versioned; the payload is not. `model.balanced_network` / `model.multiconductor_network` are serde snapshots of the live Rust IR, so tools that compute on payload fields couple to powerio internals — pio-json-schema.md already tells consumers to treat the payload as opaque or pin `producer.version`. A declared v1 payload schema is what turns `.pio.json` from a wire format into an IR contract.

Element identity is the same decision. `ElementRef { table, row }` addresses operating point updates positionally, so row assignment is an implementation detail of one parser: the PR #172 review caught uid-less GOC3 devices shifting every later device's updates onto the wrong rows. The shared `goc3_bridge` row authority fixes powerio-pkg, but any third party producer of `operating_points` must still reproduce parser row order. `source_uid` is carried but advisory — payload rows have no identity field to resolve it against.

Scope:

- Declare and version the payload field schema for both model families, and decide how payload versioning relates to the envelope `schema_version` (currently envelope only).
- Stable row identity in the payload (source uid where the format has one, synthesized otherwise); `ElementRef.source_uid` becomes authoritative when present, `row` a fallback and consistency check.
- Validation rejects a package whose updates reference unknown identities.

samtalki:
Closes #173.

The `.pio.json` payload is now a declared contract. Two optional envelope fields, `payload_schema` and `payload_schema_version`, name and version the IR payload per model kind (`pio-payload-balanced/1`, `pio-payload-multiconductor/1`, both `1.0.0`), independent of the envelope `schema_version` (now `0.1.1`): the envelope versions the package bookkeeping, the payload versions the network tables a consumer computes on. A reader rejects a foreign payload major before touching payload fields. The JSON shape of `model` is unchanged and packages written by 0.5.0 read as before.

Payload rows get stable identity. Balanced IR elements gain `uid: Option<String>` (serde additive; absent keys omitted, so non-GOC3 snapshots are byte identical). The GOC3 parser keeps source uids on buses, dispatchable devices, branches, and dc lines instead of dropping them; package construction synthesizes `{table}:{row}` uids for rows without one.

Operating point updates resolve by identity. When the referenced table carries uids, `ElementRef.source_uid` is authoritative: it selects the row, a present `row` must agree with it, and unknown or duplicated identities are rejected both at materialization and by validation (`VALIDATE.PACKAGE.OPERATING_IDENTITY`, surfaced through `pio_package_validate`). `row` may be omitted on the wire (`ElementRef::by_source_uid`). Tables without uids keep row-only semantics, which is what keeps every existing package materializing unchanged. Provenance cleanup paths now come from the resolved row rather than the wire row, and an update may not overwrite `uid`.

Surfaces: no C ABI signature changes and no `PIO_ABI_VERSION` bump; identity failures return `NULL` with the message in `errbuf`. Python table dicts expose `uid` and `Package.materialize_operating_point` raises `ValueError` on unknown identities. PowerIO.jl needs only the routine artifact repin at release.

Tests: identity resolution matrix (uid-only update with `row` omitted, row/uid mismatch, unknown uid, duplicate uids, uid overwrite, legacy uid-less payload fallback), payload version gate (same major accepted, foreign major and invalid semver rejected, absent accepted), GOC3 uid-less row shift extended to prove identity lookup rather than extractor row alignment, a C ABI materialize failure test, Python shape and error tests, and a snapshot lock that `uid` stays off the wire when absent. 886 Rust and 91 Python tests pass; clippy and fmt clean; cbindgen header unchanged.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -148,6 +148,22 @@ jobs:
         </dl>
         <footer><a href="https://github.com/eigenergy/powerio">github.com/eigenergy/powerio</a></footer>
         HTML
+    # The schema URLs stamped into .pio.json files (`schema`, `payload_schema`)
+    # are identifiers first, but the site serves them so they also resolve:
+    # each lands on the schema guide chapter. Relative targets keep the pages
+    # working under both powerio.dev and eigenergy.github.io/powerio.
+    - name: Schema identifier pages
+      run: |
+        for id in pio-package/0.1 pio-payload-balanced/1 pio-payload-multiconductor/1; do
+          mkdir -p "target/doc/schema/$id"
+          cat > "target/doc/schema/$id/index.html" <<HTML
+        <!doctype html>
+        <meta charset="utf-8">
+        <meta http-equiv="refresh" content="0; url=../../../guide/pio-json-schema.html">
+        <title>PowerIO schema: $id</title>
+        <a href="../../../guide/pio-json-schema.html">PowerIO schema $id — see the .pio.json schema guide</a>
+        HTML
+        done
     - uses: actions/upload-pages-artifact@v5
       with:
         path: target/doc
--- a/.github/workflows/julia-binding.yml
+++ b/.github/workflows/julia-binding.yml
@@ -6,6 +6,13 @@ name: Julia binding
 # it is public; while it is private, add a token with read access to it as the
 # POWERIO_JL_REPO_TOKEN secret. With neither, the job prints a skip notice and
 # passes, so missing external access never blocks a PR.
+#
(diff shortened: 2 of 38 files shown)


samtalki:
Tandem CI note: the `PowerIO.jl against this C ABI` failure was the byte equality assertion on `schema_version` in PowerIO.jl main. The job now checks out a same-named PowerIO.jl companion branch when one exists (falls back to main); eigenergy/PowerIO.jl#52 carries the constant bump and relaxes the assertion to the same-major reader contract. Either PR can merge first.

PR #234, opened 2026-07-05
[codex] Add Arrow catalog and matrix axes

### LINKED ISSUE #230: C ABI: make matrix Arrow tables self describing with axis maps
Release target: 0.6.3

Keep the existing balanced matrix Arrow tables as COO tables and add explicit axis map tables so language bindings do not need to infer row identity through normalized solver tables.

## Scope

- Keep existing matrix tables `ybus`, `incidence`, `bprime`, and `bdoubleprime` as Arrow COO tables.
- Add append only Arrow table ids for matrix axes:
  - `matrix_bus`: `index`, `bus_id`, `source_row`, `is_reference`, `component`
  - `matrix_branch`: `index`, `source_row`, `from_bus_id`, `to_bus_id`
- Add matrix schema metadata:
  - `powerio.table`
  - `powerio.format = "coo"`
  - `powerio.row_axis`
  - `powerio.col_axis`
  - `powerio.row_count`
  - `powerio.col_count`
- Use `row_axis = "matrix_bus"` and `col_axis = "matrix_bus"` for bus by bus matrices.
- Use `row_axis = "matrix_bus"` and `col_axis = "matrix_branch"` for incidence.
- Leave `solver_bus` untouched for existing low level consumers.

## Acceptance

- C header constants are append only.
- Golden Arrow fixtures include axis metadata.
- Julia can build `idx_to_bus` without `:solver_bus`.
- No ABI version bump; consumers probe table availability.

### LINKED ISSUE #231: Matrix ABI: document Arrow COO as v1 and defer CSC/CSR C buffers
Release target: 0.6.3

Document the matrix ABI boundary before adding more matrix surfaces. Arrow COO plus explicit axis maps is the v1 language neutral transport. Native sparse objects remain the language binding's responsibility.

## Scope

- Document that matrix ABI v1 is Arrow COO plus axis maps.
- State that Rust may store matrices as CSR and language bindings may assemble native sparse objects.
- Do not add `pio_matrix_csc` or raw pointer ownership APIs in this release.
- Add a follow up benchmark note: only add CSC/CSR export if COO assembly cost dominates or Arrow free consumers need it.

## Acceptance

- `powerio-capi` README explains why COO is the stable language neutral ABI.
- Docs mention Julia returns CSC because Julia owns `SparseMatrixCSC` construction.

samtalki:
## Summary

Adds a discoverable Arrow table catalog and explicit matrix axis maps to the C ABI Arrow surface.

Changes include:

- Adds `PIO_ARROW_TABLE_MATRIX_BUS = 19` and `PIO_ARROW_TABLE_MATRIX_BRANCH = 20` without renumbering existing table ids.
- Adds `pio_arrow_catalog_json()` so bindings can discover table ids, columns, axes, units, feature requirements, and schema versions.
- Adds matrix schema metadata: `powerio.schema_version`, `powerio.format`, `powerio.row_axis`, `powerio.col_axis`, `powerio.row_count`, and `powerio.col_count`.
- Updates matrix golden fixtures to include axis maps and metadata.
- Documents matrix ABI v1 as Arrow COO plus `matrix_bus` / `matrix_branch` axis maps; language bindings own native sparse matrix assembly.
- Keeps `bprime` and `bdoubleprime` as MATPOWER/PYPOWER FDPF matrices. `bprime` is `Bp`, the active power mismatch to voltage angle update matrix; it clears bus shunts and line charging, sets tap magnitudes to one, clears resistance in XB, and keeps phase shifts. `bdoubleprime` is `Bpp`, the reactive power mismatch to voltage magnitude update matrix; it clears phase shifts, clears resistance in BX, and keeps shunts, line charging, and tap magnitudes.
- Keeps DC incidence and DC OPF matrices separate as `A diag(b) A^T`; MATPOWER `makeBdc` behavior remains under `DcConvention::Matpower`.
- Expands C ABI and language docs around Arrow tables, format tokens, `.pio.json`, and release preparation for v0.6.3/v0.7.0.
- Extends the Julia benchmark harness to include parse plus matrix construction comparisons.

Closes #230.
Closes #231.

## Validation

- `cargo fmt --all --check`
- `cargo test -p powerio-matrix`
- `cargo test -p powerio-capi --features arrow,matrix --lib`
- `cargo test`

Local `python3 -m pytest python/tests` could not run because this environment does not have `pytest` installed; use the GitHub Python jobs for that signal.

## Notes

DC OPF and sensitivity Arrow exports should use a future derived product handle. The current PR keeps the `PioNetwork` table ids focused on the balanced case and matrix axis maps.


powerio-matrix/src/matrix/triplet.rs
```diff
@@ -59,13 +103,20 @@ impl CooBuilder {
     }
 
     /// Accumulate `v` into entry `(i, j)`. Skips the insert if `v == 0.0`.
+    ///
+    /// # Panics
+    /// Panics if the packed coordinate key overflows `usize`.
     #[inline]
     pub fn add(&mut self, i: usize, j: usize, v: f64) {
```
samtalki, line 68: `CooBuilder::add` packs `(i, j)` into a single `usize` key (`i*cols+j`) guarded only by `debug_assert!`, and `finish_csr` recomputes `(i, j)` from that key by division/modulo. This repo's `[profile.release]` has no `debug-assertions = true` override, so the assert compiles out in release builds. Before this change, an out-of-range column index reached `sprs`'s always-on `assert!(col < cols)` in `add_triplet` and panicked reliably in every profile. Now `j = key % cols` is in-bounds by construction, so any future/latent bug that hands this a column index `>= cols` (mismatched branch count feeding incidence/Ybus/Bp/Bpp) silently aliases into a different valid matrix cell instead of panicking — a numerically wrong matrix gets exported with no signal anything went wrong.

powerio-matrix/src/matrix/bprime.rs
```diff
@@ -1,74 +1,38 @@
...
 
+use crate::Result;
 use crate::indexed::IndexedNetwork;
-use crate::{Error, Result};
 
-use super::{BuildOptions, Scheme, triplet::CooBuilder};
+use super::ybus::{YbusFlags, build_ybus_with_flags};
+use super::{BuildOptions, Scheme, negate_into};
 
 pub fn build_bprime(case: &IndexedNetwork, opts: &BuildOptions) -> Result<CsMat<f64>> {
```
samtalki, line 27: This rewrite drops the explicit self-loop skip the old `build_bprime` had (`if i == j { /* self-loop: contributes only as a shunt, has no place in B' */ continue; }`). It now delegates to `build_ybus_with_flags`, whose `i == j` branch folds all four terminal admittance terms onto the diagonal for a self-loop and scatters that into both `g_coo`/`b_coo` — i.e. into `Y_bus`, and from there into `Bp` via `-Im(Y_bus)`. For a self-loop branch with nonzero impedance or a nonzero phase shift, `Bp`'s diagonal at that bus now picks up a contribution the old code always zeroed out, changing the FDPF P-step Jacobian for that bus. `self_loop_with_zero_reactance_drops_unconditionally` in tests.rs uses `r=x=0` and only calls `build_incidence`, so it can't catch this — both before and after this diff, `denom==0` short-circuits before the self-loop combine logic runs. Worth a dedicated nonzero-impedance self-loop test against `build_bprime`/`build_bdoubleprime` if this drop wasn't intentional.

samtalki:
**Correctness: `sddm_check` never verifies symmetry, and this PR makes `Bp` asymmetric by design**

`sddm_check` (`powerio-matrix/src/matrix/mod.rs:224`, unchanged by this diff) only checks `m_matrix_sign`, diagonal dominance, and positive diagonal — never symmetry — despite its doc comment defining SDDM as "symmetric diagonally dominant M-matrix."

This PR's `bprime.rs` rewrite keeps phase shifts in `Bp` (per MATPOWER `makeB.m`), which makes `Bp` legitimately asymmetric for cases with phase-shifting transformers — and this PR's own pegase fixture test now asserts `not is_symmetric(bprime)`. So a case with phase shifters can get `sddm: true` written into the exported `_meta.json` (via `sddm_check` in pipeline.rs) even though `Bp` isn't symmetric. A downstream consumer that trusts `sddm=true` as license to run a Cholesky-based SDDM solver on the exported matrix gets silently wrong results or a factorization failure.


**Fixed:** 
`Bp` can now be possibly asymmetric, and `sddm_check` correctly tests symmetry.

powerio-matrix/src/matrix/triplet.rs
```diff
@@ -7,14 +7,55 @@
 //! flow map, and generator→bus matrices.
 
 use std::collections::HashMap;
+use std::hash::{BuildHasherDefault, Hasher};
 
 use sprs::{CsMat, TriMat};
 
+type CoordinateMap = HashMap<usize, f64, BuildHasherDefault<CoordinateHasher>>;
```
samtalki: `CoordinateMap` swaps the standard `(usize, usize)`-keyed `HashMap` for a hand-rolled 64-bit hasher (`CoordinateHasher`: a SplitMix64-style `mix`/`write_usize`, plus a separate FNV-like byte path in `write()` that's never actually exercised by this usage). This is unaudited, untested against adversarial collision patterns, and adds ~45 lines of bit-twiddling every future maintainer of the sparse-matrix hot path has to re-verify by hand. `rustc-hash`/`FxHashMap` is the standard fix for this exact "SipHash is too slow for small integer keys" problem and would give the same speedup with none of the new code to get wrong.

powerio-capi/src/arrow_export.rs
```diff
@@ -676,51 +865,225 @@ macro_rules! real_matrix_batch {
...
+    .map_err(|e| e.to_string())
+}
+
+#[cfg(not(feature = "matrix"))]
+fn matrix_bus_batch(_net: &Network, _core: &IndexCore) -> Result<RecordBatch, String> {
+    Err(matrix_feature_error())
+}
+
+#[cfg(feature = "matrix")]
+fn matrix_branch_batch(net: &Network, core: &IndexCore) -> Result<RecordBatch, String> {
```
samtalki, line 934: This re-implements `build_incidence`'s column filter (`i == j || br.x == 0.0`) by hand instead of reusing `IncidenceParts::branch_of_col`, which `build_incidence` already computes for exactly this purpose. `docs/src/capi-arrow.md` documents `matrix_branch` as tracking the incidence matrix's columns 1:1 ("branches that do not contribute an incidence column ... are not on this axis"), but the two column sets are derived by two independent, hand-copied filters ~250 lines apart, and this one doesn't honor `opts.skip_zero_impedance`. If `build_incidence`'s filtering ever gains a case without the same edit landing here, `matrix_branch` silently stops matching the incidence column count/order — and `matrix_arrow_golden_fixtures_match` only exercises `BuildOptions::default()` on both paths, so nothing would catch it.

powerio-capi/src/lib.rs
```diff
@@ -686,6 +690,96 @@ pub unsafe extern "C" fn pio_base_mva(net: *const PioNetwork) -> f64 {
...
+) -> *mut c_char {
+    unsafe {
+        finish_string(
+            errbuf,
+            errlen,
+            "panic while serializing summary JSON",
+            || {
+                let c = network_ref(net).ok_or_else(|| "network handle is NULL".to_string())?;
+                let v = IndexedNetwork::with_core(&c.net, &c.core);
+                let reference_bus_indices = v.reference_bus_indices();
```
samtalki, line 746: Resolves `v.reference_bus_indices()` back to external bus ids via `c.net.buses.get(idx)` instead of `IndexedNetwork::bus_id(idx)` (`powerio/src/indexed.rs:226-228`), the accessor built for exactly this dense-index-to-id mapping. `IndexedNetwork::with_core` star-lowers 3-winding transformers by appending synthetic star buses to its own (possibly expanded) bus list, so `v.net.buses` can differ from `c.net.buses`. This only stays correct today because `expand_transformers_3w` appends new buses after the original ones and a reference bus is never synthetic — an implementation detail of a different module, not something this call site enforces. A future change to star-lowering order would silently return the wrong external bus id in the summary JSON with no compiler error and no fixture to catch it (none combine a 3-winding transformer with reordered buses). Prefer `v.bus_id(idx).0`.

PR #258, opened 2026-07-13
Add DeepMind OPFData JSON reader and conversion support

### LINKED ISSUE #257: Add DeepMind OPFData JSON reader and cross-format conversion
## Problem

PowerIO does not currently recognize or parse raw examples from DeepMind's OPFData dataset. A raw OPFData example is a framework-independent JSON document with `grid`, `solution`, and `metadata` sections. PyTorch Geometric reads those JSON files and converts them into `HeteroData`; its processed `.pt` files are derived caches rather than the source format.

References:

- [OPFData paper](https://arxiv.org/pdf/2406.07234)
- [PyTorch Geometric OPFDataset loader](https://github.com/pyg-team/pytorch_geometric/blob/2.6.1/torch_geometric/datasets/opf.py)
- [Official dataset README](https://storage.googleapis.com/gridopt-dataset/README)
- [Dataset license (CC BY 4.0)](https://storage.googleapis.com/gridopt-dataset/LICENSE)

## Proposed scope

Add a read-only OPFData JSON adapter in the lightweight `powerio` crate:

- Parse one extracted `example_N.json` at a time.
- Auto-detect the top-level `grid`, `solution`, and `metadata` markers.
- Add the canonical token `opfdata-json` and aliases `opfdata`, `gridopt-json`, and `gridopt`.
- Route through the shared Rust format hub so the CLI, Python bindings, and C ABI inherit support.
- Retain the raw JSON for byte-exact same-format echo.
- Convert the solved operating point through the canonical `Network` model to existing writable formats.
- Reject canonical OPFData output from arbitrary networks because this adapter is reader-only.

## Mapping and fidelity

- Synthesize one-based bus IDs from zero-based row indices.
- Read bus types, base voltage, and voltage limits from `grid`; use solved voltage magnitude and angle from `solution`.
- Resolve generators, loads, and shunts through their link tables.
- Use solved generator dispatch and terminal-bus voltage; convert p.u. powers to MW/MVAr.
- Convert quadratic generator costs from the per-unit-power basis to the canonical MW basis.
- Map AC lines and transformers into branches, preserving impedance, terminal charging, ratings, tap, shift, and solved terminal flows.
- Reorder solution flow columns from `[pt, qt, pf, qf]` into the canonical branch solution fields.
- Validate row widths, counts, link completeness and uniqueness, endpoint ranges, bus types, edge alignment, and positive finite `baseMVA`.
- Recompute `metadata.objective` from the solved dispatch and costs and warn on a material mismatch.
- Report that initial generator values are replaced by solved values, and that original bus IDs/names, areas/zones, and base frequency are absent and must be synthesized or defaulted.

## Out of scope

- PyTorch/PyG `.pt` files
- Direct `.tar.gz` archive reading
- Dataset downloading
- Batch-directory iteration
- A canonical OPFData writer

## Acceptance criteria

- [ ] An official case-14 OPFData JSON example parses with the expected element counts and representative values.
- [ ] FullTop and N-1 examples are accepted without fixed element-count assumptions.
- [ ] JSON auto-detection and all documented aliases work.
- [ ] Same-format output is byte exact.
- [ ] Conversion to MATPOWER and PowerModels JSON preserves the representable solved snapshot and reports fidelity losses.
- [ ] Malformed feature widths, counts, links, endpoints, bus types, topology alignment, and `baseMVA` fail with descriptive errors.
- [ ] CLI, Python, and C ABI paths are covered through the shared format hub.
- [ ] The official fixture records its exact source, CC BY 4.0 license, byte/line counts, and checksum.

cameronkhanpour:
## Summary

- add a read only DeepMind OPFData JSON adapter to the format neutral `BalancedNetwork` hub
- map solved bus, generator, load, shunt, line, transformer, cost, and branch flow data into canonical units
- validate feature widths, row alignment, topology, links, endpoints, bus types, base MVA, and objective consistency
- support automatic JSON detection and the `opfdata-json`, `deepmind-opfdata-json`, `opfdata`, `gridopt-json`, and `gridopt` names
- expose the shared reader and conversion path through the CLI, Python package, and C API
- retain byte exact source echo and report cross format fidelity limits

## Scope

The reader accepts one extracted raw FullTop or N-1 document at a time. Element counts and topology come from each document, so the implementation does not depend on a grid size, case name, or PGLib registry. A document from another producer must follow the published object layout, feature column order, units, and link rules. PyTorch `.pt` caches, archives, downloads, and batch directory ingestion remain outside this change.

The included case 14 fixture is the smallest published document containing every table. It is used by the Rust reader tests, CLI tests, C API tests, Python tests, and the conversion matrix. Additional validation covered FullTop examples from all ten published grid families, N-1 generator and branch outage examples, and a 500 bus conversion through MATPOWER and PowerModels JSON.

The format contract is documented in Appendix A of the OPFData paper. The PyTorch Geometric loader is the executable reference for the same layout. The upstream project does not publish a separate JSON Schema or a format version marker.

## Validation

- `cargo test`
- `cargo test -p powerio-cli`
- `cargo test -p powerio-capi`
- `maturin develop && pytest python/tests` with 107 passed and 7 skipped for optional dependencies
- `cargo fmt --all --check`
- `bash scripts/ci-clippy.sh`
- conversion matrix coverage across all eight writable transmission formats
- official fixture size and SHA-256 verification

Closes #257


docs/src/format-fidelity.md
```diff
@@ -183,6 +184,24 @@ code `READ.TRANSMISSION.PARSE_WARNING`. GridFM package reads use
   writer emits a canonical Surge network body for the supported power flow core;
   richer MATPOWER generator capability or ramp columns and unsupported cost
   shapes are reported in `Conversion::warnings`.
+- **OPFData JSON** reads one raw solved-example JSON file into the balanced
```
samtalki: "solved-example JSON file" is a little unclear, this could be any of the existing JSON formats.

docs/src/format-fidelity.md
```diff
@@ -25,6 +25,7 @@ implementations and the matching powerio code:
 | PSLF shunts | EPC `pu_mw`/`pu_mvar` are per unit on `sbase`; `Network::Shunt` stores MW/MVAr at \\(V = 1\\) | paired EPC/RAW case checks | `format::pslf` |
 | GO Challenge 3 time series | `Network` stores the first interval as a static case; `.pio.json` documents carry replayable later intervals in `operating_points` | Rust GOC3 package tests | `format::goc3`, `powerio_pkg::operating` |
 | Surge angles | Surge JSON carries voltage angles, phase shifts, and angle limits in radians; `Network` stores degrees | Rust Surge round trip tests | `format::surge` |
+| OPFData solved examples | OPFData carries p.u. powers and radian angles; `Network` stores the solved snapshot in MW/MVAr and degrees, with zero based links mapped to one based bus IDs | Smallest schema-complete official fixture plus size independent row, topology, FullTop, and N-1 contract tests | `format::opfdata` |
```
samtalki: I am worried that using the term `OPFData` directly may confuse users who use the `powerio-prob` crate to generate OPF problem data instances for downstream solvers. Is there any other term that is used for this format? 

Also, `Network` is deprecated in favor of `BalancedNetwork` and the `Network` alias will be retired in v1.0.0, so it's worthwhile to just use `BalancedNetwork` now on new code.

docs/src/python.md
```diff
@@ -23,12 +23,16 @@ pandas, or pyarrow.
...
 Transmission text and file format names accepted by `parse_*` and `convert_*` include
 `matpower`, `psse`, `powerworld`, `pslf`, `powermodels-json`, `egret-json`,
-`pandapower-json`, `goc3-json`, `surge-json`, and `powerio-json`, plus their
+`pandapower-json`, `goc3-json`, `surge-json`, `opfdata-json`, and `powerio-json`, plus their
 documented aliases. PyPSA CSV folders and GridFM Parquet datasets are directory
 formats; use `read_pypsa_csv_folder`, `Network.write_pypsa_csv_folder`,
 `read_gridfm`, `Network.write_gridfm`, or the conversion/package helpers that
 take a path.
 
+`opfdata-json` accepts one extracted raw example from any published FullTop or
```
samtalki: slightly strange placement in the docs, getting a lot of lines for this

docs/src/README.md
```diff
@@ -20,7 +20,8 @@ Public conventions:
 
 Transmission readers cover MATPOWER, PSS/E revisions 33 through 35,
 PowerWorld AUX and PWB, PSLF EPC, PowerModels JSON, egret JSON, pandapower JSON,
-PyPSA CSV folders, GO Challenge 3 JSON, Surge JSON, and GridFM Parquet
+PyPSA CSV folders, GO Challenge 3 JSON, Surge JSON, DeepMind OPFData FullTop
```
samtalki, line 23: re: my other comments about the naming, referencing deep mind explicitly seems like a good convention. `DeepMindOpfData` or `DmOpfData` seem more ergonomic but admittedly clunky/somewhat long.

powerio-capi/src/lib.rs
```diff
@@ -2473,6 +2473,37 @@ mod tests {
...
+                CStr::from_ptr(source_format.as_ptr()).to_str().unwrap(),
+                "OpfDataJson"
+            );
+            assert_eq!(len, "OpfDataJson".len());
+            assert!(warning_text(net).contains("solver initial values"));
+            assert!(to_format(net, "matpower").contains("mpc.bus"));
+            pio_network_free(net);
+        }
+    }
+
```
samtalki, line 2758: what does this test check for in simple terms? also probably needs a docstring.

tests/data/opfdataset/example_0.json
samtalki, line 1: please confirm if you want this fixture; in general I think we should adopt the policy of using fixtures in the PR conversion matrix (and several tests) if it's going to be included.

samtalki (changes requested):
@cameronkhanpour thank you this is great.
- can you try to talk with downstream users to make sure that this meets the needs of their groups so that we can avoid follow up PRs
- can you confirm that there will be no conflicts when you rebase this on to the head of #256 , and preemptively fix any conflicts on this PR now? that way, this can ride along for the v0.7.1 release.
- my main structural comment is about the nature of the deep mind opf datasets. from https://pytorch-geometric.readthedocs.io/en/latest/generated/torch_geometric.datasets.OPFDataset.html#torch_geometric.datasets.OPFDataset it looks like this only supports variants of pglib. can the format be applied more broadly outside of pglib?

samtalki:
Also, @cameronkhanpour is there an official definition/schema of this format available anywhere that we can use as a source for ground truth?

docs/src/format-fidelity.md
cameronkhanpour: I changed this to say that the reader accepts one raw JSON document from a DeepMind OPFData FullTop or N-1 release. The section now names the source before it describes how the data maps into the balanced model.

docs/src/format-fidelity.md
cameronkhanpour: I changed the enum names to `DeepMindOpfDataJson` and the docs now say `DeepMind OPFData JSON`. The public token remains `opfdata-json` because OPFData is the published dataset name. I also changed the new reader and its tests to use `BalancedNetwork`.

docs/src/python.md
cameronkhanpour: I removed the separate paragraph and folded the important point into the existing format list. It now takes one short sentence.

docs/src/README.md
cameronkhanpour: I adopted the explicit DeepMind name in the enum, source format, labels, and documentation. The format token remains `opfdata-json`, and `deepmind-opfdata-json` is also available as an alias.

powerio-capi/src/lib.rs
cameronkhanpour: This test checks that the normal C entry points can detect the file, expose its basic counts and base MVA, return its read warnings, and convert it to MATPOWER. I added that description above the test and gave the test a simpler name.

tests/data/opfdataset/example_0.json
cameronkhanpour: I want to keep the fixture because it is the smallest official file that covers every published table and it is needed for the exact source echo check. It is now used by the Rust reader tests, CLI tests, C API tests, Python tests, and the PR conversion matrix. The fixture README records its source, license, size, and hash.

cameronkhanpour:
PR #258 is now rebased onto commit `cd9fbae`, which is the current head of #256. There was one overlap in the public reexport list. I kept both sets of exports. The rebased branch now contains that commit as an exact ancestor.

The published OPFData releases are based on PGLib-OPF. The parser does not use PGLib case names or fixed element counts. It can read a document from another source when that document uses the same object layout, feature column order, units, and link rules. The docs now state this limit clearly.

The paper provides the published format definition in Appendix A. Listing 2 gives the JSON object layout. The following pages define each feature column, unit, and link. The PyTorch Geometric loader is the executable reference for the same layout. I did not find a separate JSON Schema or a format version marker. I added both sources and that limitation to the format and fixture documentation.

[OPFData paper](https://arxiv.org/pdf/2406.07234)

[PyTorch Geometric loader](https://github.com/pyg-team/pytorch_geometric/blob/2.6.1/torch_geometric/datasets/opf.py)

PR #260, opened 2026-07-21
cli: recursive case discovery; JSON parser hardening; 0.7.2 prep

samtalki:
## What

`batch` and the TUI browser scanned for `.m` files only, at `max_depth(2)`, and `batch` bailed with ``no `.m` files found`` on folders holding other supported formats. This change:

- Discovery recurses without a depth cap over `.m`, `.raw`, `.aux`, `.epc`, `.pwb`, `.json`, `.dss`. Extension matching is case-insensitive; hidden entries and the output directory subtree are pruned; results are sorted.
- Discovery, family inference (`infer_input_family`, moved out of `main.rs`), and loading live in one `cases` module shared by `batch` and the TUI, so the extension list exists in one place.
- Distribution inputs (`.dss`, BMOPF/PMD `.json`, auto-detected by the shared JSON shape classifier) parse to the multiconductor model and go through `powerio_pkg::lower_multiconductor_to_balanced` before the matrix pipeline. Lowering approximations and dropped fields surface as warnings.
- Scan mode skips files that fail to load, with a warning (stray `.json` cannot abort a run); an explicit single-file input keeps the hard error. A scan that loads nothing still errors.
- TUI case names are paths relative to the data directory, so `psse/case14.raw` and `case14.m` stay distinguishable; the batch job list reuses them.

## JSON parser review pass (second commit)

A review of the JSON reader family and the shared classifier, with the direct fixes applied here and the behavior-changing remainder filed as #262/#263. This commit also bumps the workspace to 0.7.2 and adds the changelog section, so the release tag can go on main once this merges.

- **BOM tolerance**: a leading UTF-8 byte order mark defeated `classify_json_text` and every text reader (`expected value at line 1 column 1`). The classifier, the transmission read funnel, the distribution parsers (including `.dss` and redirected files), the PMD/BMOPF split, `Network::from_json`, and `Package::from_json` now strip it; each strip is an itemized fidelity warning, since a same-format echo returns the text without the mark.
- **One read, one classification**: `parse_str_with_name` joins the format hub (`parse_str` plus the name-hint role the file stem plays in `parse_file`), and `cases::load_network` reads a `.json` once and hands the text straight to the typed parser. A batch scan previously read each `.json` twice and DOM-parsed it three times; on OPFData-sized files that dominated the load time.
- **Stem naming for nameless distribution cases**: `name` is optional in the BMOPF schema, and every nameless case lowered to the name `lowered-multiconductor`, so two of them in one batch overwrote each other's exports.
- **Robustness against crafted or malformed documents**: bounded allocations in the egret polynomial cost reader (a ~100-byte document could demand an arbitrarily large vector, and a `usize::MAX` exponent key indexed out of bounds) and the BMOPF matrix/winding readers (gigabyte allocations and quadratic work from small documents); the Surge float bus reference no longer saturates to `usize::MAX`.
- **Correctness**: PowerModels boolean status fields read out of service (previously `"br_status": false` read as in service), `baseMVA` must be finite and positive, padded gencost rows trim to `ncost` before the per-unit unscale, and an out-of-range cost model passes through unscaled instead of wrapping into the rescale; pandapower trafo `sn_mva = 0` no longer divides impedance by zero and string cells cannot smuggle `"inf"`/`"nan"` into numeric columns; GOC3 rejects duplicate `simple_dispatchable_device` uids instead of silently swapping time series; PMD rejects the MATHEMATICAL `data_model` instead of misreading it as ENGINEERING; BMOPF warns on a missing line `length` and keeps `meta` in extras; the `.pio.json` semver check accepts build metadata containing a hyphen.

## Notes

- Feeders the lowering preflight refuses (transformers, closed switches) skip with the structured diagnostic; extending the pass is #261, with corrected acceptance criteria (IEEE 13 additionally has single and two phase laterals, which no balanced lowering can carry).
- Pre-existing, more visible now: pipeline outputs are named by circuit name, so the same case in two formats (e.g. `ACTIVSg200.aux` + `.pwb`) overwrites its own exports in one batch run.
- `verify`, `dcopf`, and `sensitivities` remain MATPOWER-only.
- The bmopf-report example networks exercise the new path end to end: the ENWL example parses, lowers, and batches (stem-named exports); the IEEE 13 example is refused with the itemized diagnostics.

## Testing

- Unit tests for discovery (recursion, sorted order, pruning, capitalization), the extension-label sync, and the file-stem naming of a nameless BMOPF case.
- Regression tests for every reader fix: BOM classification/parse/warning at all three layers, PowerModels boolean status and `baseMVA` and cost trim/wrap, egret exponent bound, Surge float index, GOC3 duplicate uid, BMOPF matrix/winding bounds and `meta` stash, PMD MATHEMATICAL rejection, package semver suffixes.
- Integration tests: recursive batch with skip warnings, BMOPF `.json` discovery, unlowerable `.dss` skip vs explicit-input failure, empty-scan and nothing-loads errors.
- `cargo fmt --check`, `scripts/ci-clippy.sh`, and `cargo test --workspace` (1079 tests; `powerio-py` linked in CI) all clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,55 @@
 # Changelog
 
+## 0.7.2
+
+- CLI case discovery is recursive and covers every supported format (#260):
+  `powerio batch` and the TUI walk the input directory for `.m`, `.raw`,
+  `.aux`, `.epc`, `.pwb`, `.json`, and `.dss` files, matched case
+  insensitively, pruning hidden directories and the output directory. A file
+  that fails to load during a scan is skipped with a warning instead of
+  aborting the run, and the no-cases error names the supported extensions.
+  Distribution cases (`.dss`, BMOPF/PMD JSON) parse to the multiconductor
+  model and go through the explicit `lower_multiconductor_to_balanced` pass,
+  whose approximations and dropped fields surface as warnings.
+- A leading UTF-8 byte order mark no longer defeats the JSON classifier or
+  any text reader. `classify_json_text`, the transmission read funnel, the
+  distribution parsers (including `.dss` and redirected files), the PMD/BMOPF
+  JSON split, `Network::from_json`, and `Package::from_json` all strip it,
+  and the parse warnings itemize the removal (a same-format echo returns the
+  text without the mark).
+- `parse_str_with_name`: `parse_str` plus the name hint role the file stem
+  plays in `parse_file`. The CLI now reads and classifies a `.json` case
+  once and hands the text straight to the typed parser; a batch scan
+  previously read each `.json` twice and DOM-parsed it three times.
+- A nameless distribution case loaded by the CLI takes its file stem as the
+  network name, so batch exports no longer collide on the lowering's
+  `lowered-multiconductor` fallback.
+- JSON reader hardening from a review pass over the family:
+  - PowerModels: a status field written as a JSON boolean (`"br_status":
+    false`) reads out of service instead of in service; `baseMVA` must be
+    finite and positive; gencost rows padded to the MATPOWER matrix width are
+    trimmed to the declared `ncost` before the per-unit unscale; an
+    out-of-range cost model number passes through unscaled instead of
+    wrapping into the piecewise/polynomial rescale.
+  - pandapower: a trafo `sn_mva` of zero or below falls back to the system
+    base instead of dividing the impedance by zero; string cells can no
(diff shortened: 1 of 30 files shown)


samtalki:
All five Copilot findings were valid and are addressed in the latest commit: the empty-diagnostics fallback in the lowering error message, fallible `ncost` conversion plus a clamp to the coefficient length, and per-file itemization of byte order mark strips in redirected DSS files (with a regression test).

Issue #85, opened 2026-06-10
PowerIO.jl: bind the distribution C ABI


samtalki:
PR #82 added the distribution surface to the C ABI behind the `dist` cargo feature of `powerio-capi`: `pio_dist_parse_file`, `pio_dist_parse_str`, `pio_dist_warnings`, `pio_dist_to_format`, `pio_dist_convert_file`, `pio_dist_convert_str`, `pio_dist_network_free`, all additive. PowerIO.jl was deliberately left untouched.

This issue moves into v0.4.0 because the Julia distribution binding is the bridge needed for a BMOPF/OpenDSS validation loop before the compiler IR API settles.

Wiring Julia up needs:

- release tarballs built with the required distribution symbols, alongside the existing Arrow symbols;
- a Julia multiconductor handle type mirroring the C lifetimes (`pio_dist_network_free`, strings through `pio_string_free`);
- `parse_file` / `parse_str` / `convert_file` / `convert_str` / `to_format` / `warnings` APIs following the naming map in `docs/languages.md`;
- tandem CI coverage like the existing transmission round trip job;
- at least one DSS or BMOPF fixture that proves Julia can call the PowerIO distribution path.

Acceptance:

- PowerIO.jl can parse DSS, BMOPF JSON, and PMD JSON through the distribution C ABI.
- PowerIO.jl can convert a distribution input to BMOPF JSON and surface warnings.
- CI covers a distribution binding smoke test.
- The binding remains additive; existing Julia transmission users keep working.

samtalki:
addressed in #168

PR #284, opened 2026-08-07
Report source-row provenance from the normalize pass

samtalki:
## What this PR does

Adds `Network::to_normalized_with_source_rows(options)`. It returns the
normalized network together with a `NormalizeSourceRows` map: for each element
family, position `i` names the source row of the element at dense position `i`
in the `IndexedNetwork` view of that network. `None` marks an element the
pipeline synthesized, which has no source row. The map comes from the same pass
that filters, so the map and the filter rules cannot drift. The struct is
`non_exhaustive`, so later families stay additive. `to_normalized_with_options`
behavior does not change.

## Why

A consumer that resolves dense indices back to source elements had to
re-implement the filter rules. tellegen does exactly that today
(`reconstruct_ids`), with a runtime count assertion as the only guard against
drift. With this map, that function and its two error paths become a lookup.

## The star lowering

`IndexedNetwork` star-lowers every in-service 3-winding transformer, appending a
bus, the star branches, and a magnetizing shunt. A first version of this PR
returned `Vec<usize>` sized to the network *before* that lowering, so
`rows.buses[dense_index]` ran past the end of the vector for any case with a 3W
transformer — which PSS/E `.raw` and PSLF `.epc` both produce routinely. On the
PSLF 3W fixture the map held 3 buses against a view of 4, and 0 branches against
a view of 3.

The fields are now `Vec<Option<usize>>`, each as long as the matching table of
the lowered view, with `None` for the appended elements. `transformers_3w` stays
positional over the normalized network, because the lowering consumes the
transformers themselves. `powerio/tests/normalize.rs` carries a regression test
that asserts the lengths against the view; it fails on the first version.

## One implementation, not two

`solver_tables.rs` held a second copy of the filter rules in a private
`SourceRows`, used to populate `SolverTableIndex::*_source_rows`. That type is
deleted. `SolverTableIndex` now reads the map from the pass, so the rules exist
one time in the crate. Its field types and wire form are unchanged — they were
already `Vec<Option<usize>>` over the lowered view.

`to_normalized_with_options` calls the pass directly, so only the caller that
asks for the map pays for the lowered lengths.

## Notes

- No C ABI or Python mirror in this PR. tellegen consumes the Rust crate
  directly. Bindings can follow when a consumer needs them.
- In the v0.9.0 window, we can fold the map into `NormalizedNetwork` as a field.
  That is a breaking struct change, so it waits for the breaking release.

## Checks

`cargo test --workspace` (52 test binaries, 0 failures — the pre-existing
`solver_tables` tests assert exact source-row vectors, so they are the evidence
the unified map reproduces the old provenance), `cargo fmt --check`,
`scripts/ci-clippy.sh`, `scripts/capi-header-parity.sh`, and the conversion
matrix baseline, all clean.

## Release sequencing

Target release: **v0.8.2**. Sequence: v0.8.1 (fixes) → v0.8.2 (consumer
features) → v0.9.0 (breaking changes with deprecation shims) → v1.0.0 (clean
break, stability policy).

### DIFF
--- a/powerio/src/lib.rs
+++ b/powerio/src/lib.rs
@@ -77,7 +77,9 @@ pub use network::{
     SolverParams, SourceFormat, Storage, Switch, SwitchedShuntControl, SwitchedShuntMode,
     Transformer3W, TransformerControl, TransformerControlMode, Winding,
 };
-pub use normalize::{NormalizeOptions, NormalizedNetwork, POWER_MODELS_ANGLE_BOUND_PAD};
+pub use normalize::{
+    NormalizeOptions, NormalizeSourceRows, NormalizedNetwork, POWER_MODELS_ANGLE_BOUND_PAD,
+};
 pub use operations::Selector;
 pub use solver_tables::{
     NORMALIZED_SOLVER_TABLES_PASS, NormalizedSolverTables, SolverArcRow, SolverArcTerminal,
--- a/powerio/src/network.rs
+++ b/powerio/src/network.rs
@@ -1510,6 +1510,17 @@ fn repair_vg(vg: f64) -> Option<f64> {
     (!vg.is_finite() || vg <= 0.0).then_some(1.0)
 }
 
+/// The three element counts the star lowering changes, from
+/// [`Network::lowered_lengths`]. Every other family keeps its length, since the
+/// lowering only appends a star bus, its winding branches, and a magnetizing
+/// shunt.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub(crate) struct LoweredLengths {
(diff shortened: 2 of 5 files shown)


samtalki:
The red "PowerIO.jl against this C ABI" check is a base condition, not this diff. No companion branch with this name exists, so the job tested PowerIO.jl `main`. That branch still mirrors `.pio.json` 0.1.1, and this library rejects it ("this reader supports 0.2.x"). The one errored test is exactly the mismatch that eigenergy/PowerIO.jl#85 fixes. After #85 merges, a re-run goes green here and on every other open powerio PR that triggers this job.

---

Issue #273, opened 2026-07-31
Python Network.ptdf()/lodf() always use the uncapped dense solver


samtalki:
The plain `build_ptdf`/`build_lodf`/`build_ptdf_lodf` entry points (powerio-matrix/src/matrix/sensitivity.rs) run the dense Cholesky path unconditionally, allocating `n×n` intermediates with no size threshold. The Python bindings `Network.ptdf()`/`Network.lodf()` (powerio-py/src/lib.rs) call these directly and expose no solver/options parameter, so a Python caller cannot select the iterative CG path that `build_ptdf_lodf_with_options` already gates behind `SensitivityOptions::selected_solver_for_reduced_dimension` (default `auto`, threshold 512).

Consequence: computing PTDF/LODF from Python on a very large network forces the dense `n×n` factorization; on a case with tens of thousands of buses the intermediate buffers reach many GB and a `Vec` allocation failure aborts the host process rather than falling back to the iterative solver.

The CLI is unaffected (its `sensitivities` command defaults to `--solver auto` and routes through `for_each_ptdf_lodf_entry`). The C ABI does not expose PTDF/LODF.

Fix direction: route `Network.ptdf`/`Network.lodf` through `build_ptdf_lodf_with_options` with a default `auto` solver (and optionally accept a solver/threshold argument), so the same threshold that protects the CLI applies to the Python surface. Note this changes large-case results from exact-dense to iterative-CG above the threshold, matching CLI behavior.

Found during the reader-hardening review (#271); scoped out of that PR because it changes analysis-API solver behavior rather than reader parsing.

PR #387, opened 2026-08-20
Reuse sparse Cholesky factors for sensitivities

### LINKED ISSUE #291: Sensitivities: factor the reduced DC system once
## Current cost

The iterative sensitivity path builds only a Jacobi diagonal, then runs a new
Krylov solve for every PTDF and LODF right hand side. It also allocates scratch
vectors inside the iteration. On a large network this dominates runtime and can
perform millions of avoidable allocations.

## Required change

- Build the reference grounded internal DC factor matrix once.
- Compute one sparse factorization and reuse it for every right hand side.
- Derive each LODF right hand side from the already defined endpoint vectors.
- Preallocate solve and summation scratch buffers.
- Use a numerically stable dot product without allocating iterator results.
- Keep the internal positive graph weights distinct from the public branch
  susceptance values. Convert signs only while filling the public result.

The factor matrix is positive definite when every energized component has a
reference bus. Failure diagnostics distinguish missing references from a true
factorization failure.

## Completion evidence

- PTDF and LODF agree with an independent dense calculation on small networks;
- island and multiple reference bus cases;
- before and after factorization count, solve count, allocation count,
  allocated bytes, peak memory, and wall time;
- no allocation inside repeated triangular solves;
- the explicitly selected iterative implementation, if retained, documents
  its tolerance based numerical differences.

## 1.0 architecture alignment

Factor the reference-grounded internal DC system once for each unchanged topology and reuse solve scratch. Keep its positive factor weights private and convert to the public PowerModels branch susceptance signs only at the result boundary. Several connected components require one reference each.

cameronkhanpour:
Closes #291.

## Summary

- replace per-right-hand-side Jacobi-CG solves with one AMD-ordered sparse `LLT` factorization reused across PTDF and non-bridge LODF batches
- cap reusable RHS batches at 32 columns and approximately 8 MiB while preserving streamed output, pruning counts, reference columns, callback order, and bridge semantics
- rename the public solver surface from `iterative` to `sparse`, retire the convergence diagnostic, and add factorization failure reporting
- keep the dense oracle and the 8192-dimension/2 GiB auto-selection policy
- add `faer 0.24.4` with default features disabled and only `sparse-linalg`; public matrices remain `sprs::CsMat`

@samtalki please confirm the intentional API rename and the MIT-licensed `faer` dependency.

## Numerical comparison

Release-mode dense versus sparse comparison on case118 with pruning disabled:

- PTDF maximum absolute error: `8.327e-15`
- LODF maximum absolute error: `5.584e-14`

Last-bit differences can move values that lie at `drop_tolerance` across the pruning boundary.

## Benchmarks

Release Criterion runs use `DcConvention::ReactanceOnly` and force the compared solver.

| workload | before | after | change |
| --- | ---: | ---: | ---: |
| case118 full PTDF + LODF | 17.414 ms | 2.4022 ms | 7.25x faster |
| case2869pegase factorization, all RHS solves, and flow evaluation | 89.547 s | 192.48 ms | about 465x faster |

case118 uses 10 samples, a 1 s warm-up, and a 5 s measurement target. For case2869pegase, `drop_tolerance = f64::MAX` avoids materializing tens of millions of output coordinates while retaining all factorization, solve, and flow-evaluation work. The old solver's warm-up estimated 910 seconds for 10 samples, so its reported baseline is a Criterion quick-mode run; the sparse result is a 10-sample run with a 190.71–194.26 ms interval.

## Verification

- `cargo fmt --all --check`
- `cargo test -p powerio-matrix --lib`
- `cargo test -p powerio-matrix --test dcopf`
- focused CLI sensitivity tests
- Python binding sensitivity tests and the Python package suite
- C ABI tests
- rustdoc with warnings denied, mdBook build, and mdBook tests
- relevant clippy feature-matrix commands, including matrix + gridfm, C ABI feature sets, and the Python extension
- dependency tree confirms `faer` enables only `sparse-linalg`, `linalg`, and `sparse`, with no `faer` default, Rayon, or native-library features

## 1.0 architecture audit

Retain this factorization and reuse work for 1.0, tracked by #291. Do not merge it into a 0.9.1 patch because the public `Iterative` to `Sparse` rename is breaking. Rebuild it on the 1.0 matrix layer after the crate move in #408.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -199,12 +199,18 @@ records. Upgrade if you write text formats from names you do not control.
   extract|apply` read and classify a `.json` once instead of twice. The
   distribution reader's own document rule still applies on that path, so
   a JSON that is not a distribution case is refused as before.
-- Python `Network.ptdf()` / `lodf()` route through the auto sensitivity
-  solver (dense below the reduced-dimension threshold, iterative
-  conjugate gradient above it), matching the CLI `sensitivities`
-  command. Both take `solver="auto"|"dense"|"iterative"`. On a case above
-  the threshold results move from exact-dense to iterative-CG at a 1e-10
-  relative residual (#273).
+- DC sensitivities factor the grounded Laplacian once on the sparse path and
+  reuse its AMD-ordered Cholesky factors for batched PTDF and non-bridge LODF
+  right hand sides (#291). `SensitivitySolver::Sparse`, CLI/Python `sparse`,
+  and metadata path `sparse_cholesky` replace the iterative/CG names and the CG
+  tolerance and iteration fields are removed. `SensitivitySolveDidNotConverge`
+  is removed and its diagnostic code retired; allocation or index failures use
+  `SensitivityFactorizationFailed`. `Auto` remains dense through a
+  reduced dimension of 8192 while the predicted dense footprint is at most
+  2 GiB, then selects sparse. The sparse path requires positive finite branch
+  susceptances; dense remains available for nonsingular indefinite cases.
+  Results can differ in the last bits, so an entry at `drop_tolerance` can move
+  across the pruning boundary.
 
 ## 0.8.0
 
--- a/powerio-matrix/src/matrix/sensitivity.rs
+++ b/powerio-matrix/src/matrix/sensitivity.rs
@@ -6,9 +6,9 @@
 //! `ABA = ground_with(L, refs)`: one row/column removed per reference bus. The
 //! default public builders keep the dense Cholesky path, with dense Gaussian
 //! elimination as the nonsingular indefinite fallback. Option based builders can
-//! choose an iterative path that solves one grounded right hand side at a time
-//! and writes directly into sparse output. Disconnected networks with one
-//! reference per island are supported.
+//! choose an AMD-ordered sparse Cholesky path that factors the grounded matrix
+//! once and solves reusable batches of right hand sides. Disconnected networks
+//! with one reference per island are supported.
 //! Several references in one island are fixed angle buses; this is not a
 //! participation factor based distributed slack model.
 
@@ -16,6 +16,14 @@
 // sparse traversal read clearer than the iterator rewrites clippy suggests.
 #![allow(clippy::needless_range_loop, clippy::explicit_iter_loop)]
 
+use faer::dyn_stack::{MemBuffer, MemStack};
+use faer::linalg::cholesky::llt::factor::LltRegularization;
+use faer::sparse::linalg::cholesky::{
+    CholeskySymbolicParams, LltRef, SymbolicCholesky, SymmetricOrdering,
+    factorize_symbolic_cholesky,
+};
+use faer::sparse::{FaerError, SparseColMatRef, SymbolicSparseColMatRef};
+use faer::{Conj, MatMut, Par, Side, Spec};
 use sprs::CsMat;
 
 use crate::indexed::IndexedNetwork;
@@ -27,15 +35,13 @@ use crate::{Error, Result};
 
 /// Entries below this magnitude are dropped from the emitted sparse matrices.
 const PRUNE: f64 = 1e-12;
-const DEFAULT_CG_TOLERANCE: f64 = 1e-10;
-const DEFAULT_CG_MAX_ITERATIONS: usize = 20_000;
-/// Reduced-dimension ceiling for the `Auto` dense path. The old value of 512
-/// was far below the real crossover: at nr = 600 the dense path is a ~7e7
-/// flop factorization while the iterative path runs ~1200 conjugate-gradient
-/// solves, so `Auto` picked the slower solver by one to three orders of
-/// magnitude across the whole range that holds the common published cases
-/// (1354, 2869, 3120, 6470 buses).
+/// Reduced-dimension ceiling for the `Auto` dense oracle path. Kept at the
+/// policy established in #295; the separate footprint guard prevents a wide
+/// problem from selecting dense output allocations through this dimension
+/// check alone.
 const DEFAULT_AUTO_DENSE_THRESHOLD: usize = 8192;
+const MAX_RHS_BATCH_COLUMNS: usize = 32;
+const RHS_BATCH_MEMORY_BUDGET: usize = 8 << 20;
 
 /// Memory ceiling for the `Auto` dense path. The dimension alone does not
 /// bound the cost: the dense path also materializes an m x n PTDF and an
@@ -61,13 +67,13 @@ const LODF_ISLAND_TOLERANCE: f64 = 1e-9;
 #[serde(rename_all = "snake_case")]
 #[non_exhaustive]
 pub enum SensitivitySolver {
-    /// Dense below [`SensitivityOptions::auto_dense_threshold`], iterative above it.
+    /// Dense below [`SensitivityOptions::auto_dense_threshold`], sparse above it.
     #[default]
     Auto,
     /// Dense grounded inverse. This is the historical builder path.
     Dense,
-    /// Preconditioned conjugate gradient, one right hand side at a time.
-    Iterative,
+    /// AMD-ordered sparse Cholesky with reusable batched solves.
+    Sparse,
 }
 
 /// Solver path actually used for a sensitivity build.
@@ -77,7 +83,7 @@ pub enum SensitivitySolver {
 pub enum SensitivitySolverPath {
     DenseCholesky,
     DenseInverse,
-    IterativeCg,
+    SparseCholesky,
 }
 
 impl SensitivitySolverPath {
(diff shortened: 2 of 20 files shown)


powerio-matrix/src/matrix/sensitivity.rs
```diff
@@ -629,6 +652,164 @@ fn ensure_iterative_solver_eligible(inc: &IncidenceParts) -> Result<()> {
...
+            .factorize_numeric_llt(
+                &mut values,
+                numeric_a,
+                Side::Upper,
+                LltRegularization::default(),
+                Par::Seq,
+                MemStack::new(&mut factor_scratch),
+                Spec::default(),
+            )
+            .map_err(|_| Error::SingularNetwork)?;
```
samtalki, line 760: `faer::sparse::linalg::LltError` is two variants, `Numeric(LltError)` and `Generic(FaerError)`. Discarding it with `|_|` reports an allocation or index-overflow failure inside `factorize_numeric_llt` as `SingularNetwork` / `BUILD.SENSITIVITY.SINGULAR`, whose message is "the reference-grounded Laplacian is singular even though every component is grounded". On a large case that runs out of memory during numeric factorization the user is told their network is singular and goes looking at reference coverage. This is the one factorization path that bypasses the `SensitivityFactorizationFailed` variant the PR adds for exactly this.

```suggestion
            .map_err(|error| match error {
                faer::sparse::linalg::LltError::Numeric(_) => Error::SingularNetwork,
                faer::sparse::linalg::LltError::Generic(generic) => {
                    map_faer_error("sparse numeric factorization failed", generic)
                }
            })?;
```

powerio-matrix/src/matrix/sensitivity.rs
```diff
@@ -27,15 +35,13 @@ use crate::{Error, Result};
...
-/// was far below the real crossover: at nr = 600 the dense path is a ~7e7
-/// flop factorization while the iterative path runs ~1200 conjugate-gradient
-/// solves, so `Auto` picked the slower solver by one to three orders of
-/// magnitude across the whole range that holds the common published cases
-/// (1354, 2869, 3120, 6470 buses).
+/// Reduced-dimension ceiling for the `Auto` dense oracle path. Kept at the
+/// policy established in #295; the separate footprint guard prevents a wide
+/// problem from selecting dense output allocations through this dimension
+/// check alone.
 const DEFAULT_AUTO_DENSE_THRESHOLD: usize = 8192;
```
samtalki, line 42: Keeping the 8192 ceiling means `Auto` never selects the path this PR adds for any case that exists. Measured on this branch, release build, `powerio sensitivities case2869pegase --drop-tolerance 1e300`:

- `--solver dense` 20.78 s
- `--solver sparse` 0.25 s

`Auto` picks dense there: nr = 2868 is under 8192 and the predicted dense footprint is about 340 MB, well under the 2 GiB budget. The criterion benches on this branch put case118 at 1.80 ms dense against 1.50 ms sparse, so sparse is already ahead at 118 buses and the crossover sits below the smallest case in the corpus. #295 raised the ceiling because the CG path was one to three orders of magnitude slower than dense in that range; the direct factorization inverts that measurement, so leaving the ceiling where it is reproduces the same failure in the other direction — the default costs 83x on a published case. Either drop the threshold to something near the real crossover, or make `Auto` prefer sparse whenever the susceptances make it eligible and keep dense as the indefinite fallback.

powerio-matrix/src/matrix/sensitivity.rs
```diff
@@ -629,6 +652,164 @@ fn ensure_iterative_solver_eligible(inc: &IncidenceParts) -> Result<()> {
...
+
+        // `a` is CSR. Its transpose view is CSC over the same three slices,
+        // with no triplet rebuild or numeric copy. The grounded Laplacian is
+        // exactly symmetric, so the view describes the same operator.
+        let a_csc = a.transpose_view();
+        let indptr = a_csc.indptr();
+        let symbolic_a = SymbolicSparseColMatRef::new_checked(
+            dimension,
+            dimension,
+            indptr.raw_storage(),
```
samtalki, line 721: `IndPtrBase::raw_storage` is documented as the unchecked accessor: it hands back the storage whether or not the indptr is proper, and sprs offers `as_slice()` -> `Option<&[usize]>` for the checked form. faer does not close the gap either — its `check_col_ptr` verifies monotonicity, length `ncols + 1`, and `col_ptr[ncols] <= row_idx.len()`, but never `col_ptr[0] == 0`. Today the only caller passes an owned `ground_with` result, so this is latent; the day someone hands `factor` a `CsMatView` sliced out of a larger matrix, the offsets are wrong and faer accepts them, factoring a different matrix. `as_slice().ok_or_else(|| sensitivity_factorization_failed(...))?` costs nothing and makes the assumption enforced rather than commented.

CHANGELOG.md
```diff
@@ -195,12 +195,18 @@ records. Upgrade if you write text formats from names you do not control.
   extract|apply` read and classify a `.json` once instead of twice. The
   distribution reader's own document rule still applies on that path, so
   a JSON that is not a distribution case is refused as before.
-- Python `Network.ptdf()` / `lodf()` route through the auto sensitivity
-  solver (dense below the reduced-dimension threshold, iterative
-  conjugate gradient above it), matching the CLI `sensitivities`
-  command. Both take `solver="auto"|"dense"|"iterative"`. On a case above
-  the threshold results move from exact-dense to iterative-CG at a 1e-10
-  relative residual (#273).
+- DC sensitivities factor the grounded Laplacian once on the sparse path and
```
samtalki, line 202: This replaces the unreleased #273 entry rather than amending it, so 0.9.0 no longer records that Python `Network.ptdf()` / `lodf()` started routing through the auto sensitivity solver and gained a `solver` keyword at all. Someone upgrading from 0.8.x reads the new bullet as a rename of an API they never had. Worth keeping the #273 sentence and folding the rename into it.

samtalki (commented):
Read the whole diff and built the branch. `cargo test -p powerio-matrix --lib --test dcopf` passes, and dense/sparse agree to 1e-9 on an extra adversarial graph I built (20-bus ring with 40 interleaved radial stubs, 60 branches so bridge columns land at varied offsets inside the 32-column LODF batches) — the batching, the packed non-bridge columns, and the callback order all check out. Four comments inline; the `Auto` threshold one is the substantive one.

powerio-matrix/src/matrix/sensitivity.rs
samtalki: Retracting this one: the premise is wrong. `factorize_numeric_llt` here returns `faer::linalg::solvers::LltError`, not the two-variant `faer::sparse::linalg::LltError` I quoted, and its only variant is `NonPositivePivot { index }`. I confirmed it by annotating the closure argument and reading the type off the mismatch. There is no `Generic(FaerError)` on this path, so no allocation or index failure can arrive here and be misreported.

That also makes `SingularNetwork` the right classification: `ensure_sparse_solver_eligible` has already guaranteed positive finite susceptances, so a nonpositive pivot means the grounded matrix really is numerically singular. Nothing to change. Sorry for the noise.

powerio-matrix/src/matrix/sensitivity.rs
samtalki: Correcting my own numbers, and adding the part I missed.

The sparse figure above (0.25 s) was wrong. Re-measured on this branch, release build, `powerio sensitivities tests/data/case2869pegase.m`: 22.1 s dense, 2.2 s sparse. In process, excluding I/O, 20.9 s against 1.75 s. So 10x to 12x, not 83x. The direction and the conclusion hold.

Sweeping synthetic PEGASE-like cases for the crossover, mean ms per build:

```
   n     m   dense_ms  sparse_ms  ratio
  30    39      0.057      0.068   0.84
  40    52      0.090      0.095   0.94
  50    65      0.152      0.148   1.03
  70    92      0.358      0.305   1.18
 300   399     19.430      7.129   2.73
1200  1599   1227.311    133.618   9.19
3000  3999  29409.191   1250.378  23.52
```

Dense wins only below a reduced dimension of about 50, by tens of microseconds, and loses without bound above it.

The part I missed: lowering the ceiling alone would be a regression. `ensure_sparse_solver_eligible` refuses nonpositive or nonfinite susceptances, so every indefinite case that `Auto` handles today through the dense inverse fallback would start erroring instead. `auto_sparse_rejects_non_positive_susceptance` pins exactly that behavior with `auto_dense_threshold: 0`. The ceiling of 8192 is what hides it.

So the fix is two parts: move the default to 64, and resolve `Auto` against the case rather than its shape alone, so an `Auto` case the sparse path refuses falls back to dense. An explicit `Sparse` still reports the refusal.

Verified against the dense oracle at `drop_tolerance = 0`: case118 max |ΔPTDF| 8.3e-15, max |ΔLODF| 5.6e-14; case2869pegase 4.2e-13 and 3.6e-11. Full workspace suite green.

samtalki:
Superseded by #432, which rebuilds this design on the v1 architecture stack: faer sparse Cholesky factored once, blocked right hand sides, streamed emission preserved.

PR #388, opened 2026-08-20
Complete byte routing and DC OPF contracts

samtalki:
Adds strict UTF-8 byte classification and model JSON parsing, including BOM handling and coded malformed-input errors.

Completes the affine DC OPF contract with fixed nodal withdrawal and branch flow offsets, retains the unrated-limit synthesis choice on each instance, and emits the missing bundle vectors and manifest rows.

Validation:
- cargo fmt --all --check
- powerio, powerio-matrix, and powerio-prob tests with matrix support
- full CI clippy feature matrix
- rustdoc for all affected crates
- git diff --check


samtalki:
Superseded by #390 and #391, which keep the JSON byte API and DC OPF contract independently reviewable. The combined branch is retained.

PR #401, opened 2026-08-23
Route Auto to the sparse sensitivity path

samtalki:
Two fixes from the review of #387, on top of `ck/laplacian-factorization`.

## Auto never selects the sparse path

The `Auto` dense ceiling of 8192 was set in #295 against the conjugate gradient path this branch replaces. Against sparse Cholesky the comparison inverts. Release build, synthetic PEGASE-like cases, mean ms per build:

```
   n     m   dense_ms  sparse_ms  ratio
  30    39      0.057      0.068   0.84
  40    52      0.090      0.095   0.94
  50    65      0.152      0.148   1.03
  70    92      0.358      0.305   1.18
 300   399     19.430      7.129   2.73
1200  1599   1227.311    133.618   9.19
3000  3999  29409.191   1250.378  23.52
```

Dense wins below a reduced dimension of about 50, by tens of microseconds, and loses without bound above it. `powerio sensitivities tests/data/case2869pegase.m` is 22.1 s dense against 2.2 s sparse, so on the corpus `Auto` never reached the path #291 added.

The default moves to 64. Lowering it alone would have been a regression: `ensure_sparse_solver_eligible` refuses nonpositive and nonfinite susceptances, so indefinite cases that `Auto` answers today through the dense inverse fallback would start erroring instead, which `auto_sparse_rejects_non_positive_susceptance` pins with `auto_dense_threshold: 0`. The 8192 ceiling is what kept that hidden. `Auto` now resolves against the case and not its shape alone, and falls back to dense for those; an explicit `Sparse` still reports the refusal. That test splits into the two behaviors.

## Unchecked CSC index pointers

`IndPtrBase::raw_storage` is sprs's unchecked accessor and faer's `check_col_ptr` does not test that the pointers start at zero, so a view sliced out of a larger matrix would be accepted and factor the wrong operator. Only the owned `ground_with` result reaches it today. `as_slice` is the checked form and errors instead.

## Verification

Against the dense oracle at `drop_tolerance = 0`: case118 max |ΔPTDF| 8.3e-15 and max |ΔLODF| 5.6e-14; case2869pegase 4.2e-13 and 3.6e-11. Full workspace suite green, `cargo fmt --check` and `clippy --all-targets -D warnings` clean.

The changelog also regains the #273 entry, which the #291 bullet replaced rather than amended.

One review comment on #387 is retracted: the `LltError` on the factorization path is `faer::linalg::solvers::LltError`, whose only variant is `NonPositivePivot`, so `SingularNetwork` is the correct mapping and nothing changes there.

## 1.0 architecture audit

Retain automatic sparse routing, dense fallback, checked CSC handling, and the measured tests. Rebuild it above the retained #387 work in the 1.0 matrix stack and fold in applicable sensitivity allocation fixes from #405. Do not ship its public rename in 0.9.1.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -199,18 +199,8 @@ records. Upgrade if you write text formats from names you do not control.
   extract|apply` read and classify a `.json` once instead of twice. The
   distribution reader's own document rule still applies on that path, so
   a JSON that is not a distribution case is refused as before.
-- DC sensitivities factor the grounded Laplacian once on the sparse path and
-  reuse its AMD-ordered Cholesky factors for batched PTDF and non-bridge LODF
-  right hand sides (#291). `SensitivitySolver::Sparse`, CLI/Python `sparse`,
-  and metadata path `sparse_cholesky` replace the iterative/CG names and the CG
-  tolerance and iteration fields are removed. `SensitivitySolveDidNotConverge`
-  is removed and its diagnostic code retired; allocation or index failures use
-  `SensitivityFactorizationFailed`. `Auto` remains dense through a
-  reduced dimension of 8192 while the predicted dense footprint is at most
-  2 GiB, then selects sparse. The sparse path requires positive finite branch
-  susceptances; dense remains available for nonsingular indefinite cases.
-  Results can differ in the last bits, so an entry at `drop_tolerance` can move
-  across the pruning boundary.
+- DC sensitivities factor the grounded Laplacian once on the sparse path and reuse its AMD-ordered Cholesky factors for batched PTDF and non-bridge LODF right hand sides (#291). `SensitivitySolver::Sparse`, CLI/Python `sparse`, and metadata path `sparse_cholesky` replace the iterative/CG names, and the CG tolerance and iteration fields are removed. `SensitivitySolveDidNotConverge` is removed and its diagnostic code retired; allocation or index failures use `SensitivityFactorizationFailed`. Python `Network.ptdf()` / `lodf()` route through the auto solver and take `solver="auto"|"dense"|"sparse"`, matching the CLI `sensitivities` command (#273).
+- `SensitivityOptions::auto_dense_threshold` defaults to 64, down from the 8192 that #295 set against the conjugate gradient path. Dense stops beating the sparse Cholesky path near a reduced dimension of 50: case2869pegase is 22.1 s dense against 2.2 s sparse, and a 1200 bus case is 9x. `Auto` therefore takes sparse on every published case above a few dozen buses, and takes dense below that, when the predicted dense footprint would exceed 2 GiB, or when the case falls outside the sparse path's positive finite susceptance requirement. An explicit `Sparse` still reports that refusal as an error. Set `auto_dense_threshold` to restore the old routing. Because more cases now resolve to sparse by default, results can differ in the last bits and an entry at `drop_tolerance` can move across the pruning boundary.
 
 ## 0.8.0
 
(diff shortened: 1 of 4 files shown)


samtalki:
@ckhanpour3 — a measurement this PR's threshold change makes relevant to #387, reported rather than pushed since it is outside this diff's scope.

## Lowering `auto_dense_threshold` 8192 → 64 puts `branch_flow` on the hot path

`d6a52e8` drops `DEFAULT_AUTO_DENSE_THRESHOLD` from 8192 to 64, so essentially every real case now takes `sparse_ptdf_lodf_entries` where it previously took the dense builder. That makes the sparse fill loops the common path for the first time, and one pre-existing line in them costs more than it should.

`branch_flow` (`sensitivity.rs:862`, `976c409`, not part of your factorization work) resolves both endpoints through `Grounding::reduced` on every call:

```rust
let theta_from = g.reduced(from[branch]).map_or(0.0, |i| theta[i]);
let theta_to   = g.reduced(to[branch]).map_or(0.0, |i| theta[i]);
```

`Grounding::reduced` is a `partition_point` binary search over the grounds vector. It is called twice per `(branch, column)` pair — `2·m·n` times filling PTDF and `2·m·m` again filling LODF — for a value that depends only on `branch`. `Grounding::full_of_reduced` exists ten lines away with the docstring *"Hoists `reduced` out of a loop that would otherwise call it per column"*, so the pattern is already recognized in this file; the endpoint maps just never got the same treatment.

### Measured

Hoisting `rfrom`/`rto` as `Vec<Option<usize>>` computed once per call (4 lines, `+2·m·8` bytes), `SensitivitySolver::Sparse` forced, `b = 1/x`, min-of-5, release:

**case2869pegase** (`nr = 2868`, `m = 4582`)

| grounded buses | current | hoisted | |
|---|---|---|---|
| 1 (single slack) | 3960 ms | 4085 ms | no effect |
| 16 | 5402 ms | 4190 ms | **−22.4% (1.29×)** |
| 64 | 5544 ms | 4087 ms | **−26.3% (1.36×)** |

**case118**, 64 grounded: 1.14 ms → 0.68 ms (**1.68×**).

The signature is in the columns, not the ratios: the hoisted build is flat across slack counts (4085 / 4190 / 4087 ms) while the current one climbs with them (3960 / 5402 / 5544 ms). That is the `log k` of the binary search showing up, multiplied by `m·(n + m)`.

**Correction to a figure I gave earlier in this stack:** I had cited a single-slack speedup. It does not reproduce — with one ground the search is over a 1-element slice and the branch predicts perfectly, so there is no measurable difference. The win is entirely on multi-slack / multi-island cases, which is also where `check_reference_coverage` is doing real work.

### Not reported

I also looked at the LODF batch zeroing — `rhs[..nr * (last_outage - first_outage)].fill(0.0)` zeroes the full batch width while bridge columns consume no right-hand side, so their zeroing is pure waste. Quantified, it is bounded by `nr·m·8` bytes of `memset` over the whole loop (~105 MB on pegase, ~10 ms against a 4 s build, and only that high if every branch is a bridge). Real, but not worth a change on a meshed transmission case. It would matter on a radial feeder, where nearly every branch *is* a bridge.

Happy to push the `branch_flow` hoist as its own PR on top of this stack if you want it — it is confined to `sparse_ptdf_lodf_entries` and does not touch the factorization reuse in `b66e1e5`.

---

samtalki:
Superseded by #432: Auto routes to the sparse path there, with the dense crossover reset for a direct factorization.

PR #402, opened 2026-08-23
Report normalize source mappings over the C ABI

### LINKED ISSUE #399: C ABI: expose stable element mappings without public solver tables
## Gap

The C ABI cannot expose stable element mappings that interpret matrix and instance results unless the caller enables Arrow or extracts another network representation. Language bindings then repeat filtering and synthetic element construction, allocate another set of arrays, and can disagree with PowerIO.

The 1.0 design removes public solver preparation tables. Stable element identity still belongs on every public result.

## Required 1.0 behavior

- A matrix or instance result carries every bus, branch, generator, terminal, conductor, and other mapping needed to interpret its numerical rows and columns.
- Every mapping carries a stable module element ID. A synthetic element receives a stable synthetic ID and can have zero source spans. Source locations remain separate source map data.
- The C ABI provides ABI v6 owned result arrays, size queries and bulk fill operations, or immutable spans tied to a retained owner handle.
- The default C build exposes mappings without Arrow, JSON serialization, or another network extraction.
- PowerIO constructs each mapping during the same transformation that constructs the numerical result. A binding does not parse or extract the network again.
- Rust, C, Python, and Julia use the same names and element order.

## Completion tests

- filtering and out of service elements;
- three winding transformer expansion and synthetic buses;
- several generators on one bus;
- incidence rows that report unsupported branches;
- owner lifetime and release ordering under C sanitizers;
- allocation tests proving bulk extraction does not construct per element C objects or a second network representation.

## Ownership

Primary implementation branch: `agent/v1-abi6`. It advances this issue. Close only after Julia mapping parity passes. Companion Julia issue: eigenergy/PowerIO.jl#112.

samtalki:
Closes #399. Stacked on #401.

## The gap

`SolverTableIndex` carries the source row of every normalized bus, load, shunt, branch, switch, generator, storage and hvdc row, and 0.9 made it the single map on purpose — *"one map instead of two that could drift"*. Over the C ABI it reaches a caller two ways, neither of which suits a caller who just parsed a file: the Arrow solver tables carry a `source_row` column, but `arrow` is off by default in `powerio-capi` (`default = ["dist", "pkg"]`), so a stock build has no `pio_to_arrow` at all; and a `.pio.json` package's derived block needs a package. `grep -rn "source_row" powerio-capi/src/` outside `arrow_export.rs` returns nothing.

So a binding re-derives the drop rule outside the library. PowerDiff.jl keeps a generator if its `status` is nonzero and its bus survived, on top of a full unfiltered second extraction, and gives up with an error when the counts disagree. That is the second map the 0.9 note says should not exist, rebuilt from a guess at the rule.

## What it costs, concretely

The guess is wrong wherever normalization does something it does not model. Three-winding star lowering is the case the same note calls out as newly covered: it adds a bus and three branches that no file row produced, so neither "status == 0" nor "endpoint did not survive" reproduces them, and the consumer's length check fails a case the library handled fine. `solver_index_json_reports_no_source_row_for_a_star_bus` pins exactly that shape — 10 bus rows for 9 buses, one `null`, three `null` branch rows.

## Shape

One symbol rather than eight buffer accessors, which the issue offers as the alternative and which reads better here: the index is a handful of integer vectors, it also carries `bus_ids`, `reference_bus_indices`, `component_labels` and the arc index vectors consumers rebuild today, and it needs no new buffer protocol.

```c
char *pio_solver_index_json(const PioNetwork *net, char *errbuf, size_t errlen);
```

The document is `{"schema": "powerio.solver_index", "powerio_version", "pass", "index"}`. A row with no source row is `null` rather than a sentinel — JSON has the spelling, so the C `-1` convention buys nothing here.

One normalize pass runs per call and nothing is cached on the handle; the doc comment says so, so a caller reading several vectors reads the document once. That is still one pass against the two extractions plus a guess it replaces.

## Additive

No existing symbol changes and `PIO_ABI_VERSION` stays 5, per the additive rule.

## Verification

Three new tests: the source mapping on `t_case9_oos.m` (the dropped generator is row 1, survivors keep 0 and 2), the star-bus `null`s above, and the NULL-handle path. Full `powerio-capi` suite green (84 passed), `scripts/capi-header-parity.sh` and `scripts/capi-header-regen.sh` both clean against the regenerated `powerio.h`, `cargo fmt --check` and `clippy --all-targets -D warnings` clean.

## Note on the base

#401 needed `main` merged into it: `ck/laplacian-factorization` predates `3754066` (Rust 1.98 clippy) and #396's `pio_scopf_to_json_with_index_base`, which PowerIO.jl main now requires, so both of its CI failures were the stale base rather than its diff. This branch sits on top of that merge.

The Julia companion is eigenergy/PowerIO.jl#112.

---

## 1.0 architecture audit

Do not merge this public source-row or normalized solver-table API. Stable element mappings belong on matrix and instance results under #399 and #413; source locations remain separate. Reuse only the private mapping, normalization, and allocation work.


samtalki:
Superseded by the v1 architecture stack (#426): the C ABI surface this targeted was replaced by the v1 module and diagnostics layout in #425.

samtalki:
Correction: the stack reference above is stale (#426 was renumbered). The C ABI surface this targeted is replaced by the module and diagnostics layout in #425 and the ABI v6 handle surface in #443, which carries the mapping logic without exposing solver rows.

Issue #292, opened 2026-08-08
Guard the degenerate numerical paths in matrix assembly and sensitivities


samtalki:
Found in the release audit for v0.8.1. Each item silently produces a wrong matrix rather than an error, and each fix changes results for the degenerate input that triggers it — so they are grouped here for the breaking window.

The `baseMVA <= 0` case from the same audit is already fixed in 0.8.x (a zero base put NaN on every Ybus diagonal and shipped it); these are the remaining ones.

## 1. Unguarded tap divisions

`powerio-matrix/src/matrix/ybus.rs` (~`178-193`)

`effective_tap()` only remaps exact `0.0 → 1.0`. `tap = 1e-200` underflows `a_norm_sqr` to zero and scatters `±Inf`/`NaN` into Y_bus; `tap = 1e-8` scales `y_ff` by 1e16 and destroys conditioning with no diagnostic. The function rejects a non-finite `r²+x²` but never checks its four outputs. Nothing upstream constrains tap (`validate_values`/`repair` cover `vm`, `va`, `mbase`, `vg` only).

Fix: require `tap.is_finite() && tap.abs() > f64::MIN_POSITIVE.sqrt()`, and `is_finite()` the four returned admittances.

## 2. LODF denominator: cancellation behind an absolute tolerance

`powerio-matrix/src/matrix/sensitivity.rs` (~`33`, `370-372`, `512-514`)

`denom = 1.0 − delta(k,k)` with an absolute islanding test `< 1e-9`. A true bridge is caught, but a *near*-bridge (`delta(k,k) = 1 − 1.1e-9`) passes and carries ~1e-7 relative error, so the whole LODF column is amplified to ~1e9 with ~7 digits lost — and the `drop_tolerance` filter does not catch it, because the entries are huge, not tiny.

Fix: decide islanding topologically (a branch redistributes nothing iff it is a bridge of the in-service graph — an O(n+m) DFS; `IndexedNetwork` already carries petgraph and union-find), zero those columns structurally, and keep the tolerance only as a *relative* backstop that reports a diagnostic when it fires.

## 3. Pivot acceptance is absolute, not scale-relative

`sensitivity.rs` (~`801-826` Cholesky, `~639` Gauss-Jordan)

[...]

PR #403, opened 2026-08-23
Carry the DC incidence parts over the C ABI, with their guards

### LINKED ISSUE #400: C ABI: expose DC incidence, branch susceptance, and phase shift data
## Gap

The C ABI exposes assembled square matrices but not the DC data needed to build or differentiate another formulation:

- incidence matrix in branch by bus orientation;
- branch susceptance in incidence row order;
- phase shift bus injection;
- stable `ElementId` mapping for every included incidence row;
- stable `ElementId` and a diagnostic reason for any branch the requested operation cannot represent;
- selected branch susceptance formula.

Consumers currently recompute these values from raw branch data. That duplicates formula selection, numerical guards, row filtering, and element mapping outside PowerIO.

## Required semantics

The public data must match PowerModels:

```text
A[e, from] = +1
A[e, to]   = -1
B  = A' * Diagonal(b) * A
Bf = Diagonal(b) * A
p_shift = A' * (b .* shift)
p_bus = -B * va + p_shift
p_branch = -Bf * va - b .* shift
```

The branch susceptance formulas are:

```text
series susceptance      imag(inv(r + im*x))
tap adjusted reactance  -1/(x*tap)
reactance only          -1/x
```

Do not encode a universal sign assumption. Negative reactance remains valid when the selected formula is defined. Zero impedance branches remain present by default. Only an explicit checked transformation can remove one.

## C ABI shape

Return owned ABI v6 result arrays or provide size queries and bulk fill operations for incidence entries, branch susceptance, phase shift injection, included and omitted stable element mappings, and the selected formula. Source spans remain in the module source map and are not durable row identity. Sign and unit conversion happens while filling output buffers without another temporary vector.

The Arrow matrix axis tables can append the same values, but the core DC data remains available without the Arrow feature.

## Completion tests

- incidence orientation and endpoint signs;
- each branch susceptance formula;
- direct equality with `A' * Diagonal(b) * A` and `Diagonal(b) * A`;
- bus injections and affine branch flows;
- phase shifts remain separate from the symmetric bus susceptance matrix;
- included and omitted stable element mappings;
- zero impedance preservation and explicit transformation behavior;
- a negative reactance branch;
- C, Python, and Julia extraction uses one bulk allocation per returned array.

## Ownership

Primary implementation branch: `agent/v1-abi6`. It advances this issue. Close only after Julia extraction passes. Companion Julia issue: eigenergy/PowerIO.jl#114. Matrix semantics originate on `agent/v1-dc-matrices`.

samtalki:
Closes #400. Stacked on #402, which is stacked on #401.

## The gap

`IncidenceParts` holds `a`, `b`, `p_shift`, `branch_of_col` and `skipped_zero_impedance`. Only `a` reaches a caller, through `PIO_ARROW_TABLE_INCIDENCE`. `PIO_ARROW_TABLE_MATRIX_BRANCH` is `(index, source_row, from_bus_id, to_bus_id)` — no susceptance column — and `p_shift` and the skips have no representation at all. Everything else about the DC system is available only pre-summed, as `PIO_ARROW_TABLE_BPRIME` and friends.

## Why the vector, not the matrix

A differentiable-modeling consumer treats the per-branch susceptance as a **parameter vector**. PowerDiff.jl differentiates DC OPF solutions with respect to `b` and to per-branch switching `sw`, so its DC system is `L = A * diag(b .* sw) * A'`, rebuilt as those parameters move. An assembled `B'` has already summed away the thing being differentiated, so the matrix API cannot serve it at any granularity.

With no way to ask, it computes the vector itself — `z2 = r^2 + x^2; b = z2 > 1e-10 ? -x/z2 : 0.0`. The formula agrees with `SeriesImpedance` negated. **The guards do not**, and that is the part that matters. #292 hardened this exact path:

- the denominator is bounded on magnitude (`MIN_DIVISIBLE_MAGNITUDE`) rather than compared against a fixed constant, so `x = 1e-300` no longer annihilates a diagonal;
- a non-finite susceptance is `NonFiniteSusceptance` rather than a silently zero-weight edge;
- a `Matpower` tap that cannot be divided by is `DegenerateTap`, checked after the products are formed.

A consumer reimplementing the formula gets none of it, with no way to inherit it short of assembling a matrix it does not want.

## Shape

Four extractors in the existing cap/count idiom, behind `--features matrix`:

```c
ptrdiff_t pio_branch_susceptance(const PioNetwork *net, const char *convention,
                                 double *out, size_t cap, char *errbuf, size_t errlen);
ptrdiff_t pio_phase_shift_injection(const PioNetwork *net, const char *convention,
                                    double *out, size_t cap, char *errbuf, size_t errlen);
ptrdiff_t pio_incidence_branch_rows(const PioNetwork *net, const char *convention,
                                    int64_t *out, size_t cap, char *errbuf, size_t errlen);
ptrdiff_t pio_incidence_skipped_rows(const PioNetwork *net, const char *convention,
                                     int64_t *out, size_t cap, char *errbuf, size_t errlen);
```

Three choices worth stating:

- **`ptrdiff_t` with an `errbuf`, not `size_t`.** The guard's refusal *is* the value here — inheriting it is the whole point — and a plain `0` return would throw the message away. This is `pio_scenario_ids`' shape.
- **The column map and the skips travel with the vectors.** `branch_of_col` is the only way to line `b` up with the case's branch table, and a consumer that rebuilds `B` from `A` and `b` needs to know which branches the builder dropped, or its matrix and the library's disagree with nothing to say why.
- **Buffer accessors rather than the additive Arrow columns.** The issue offers both. Arrow leaves the vectors behind a feature that is off by default; these need `matrix` alone. The Arrow columns stay available additively later if the catalog wants them.

`convention` takes the tokens Python's `convention=` and the CLI's `--convention` already accept, case and separator insensitive, with NULL meaning the default `series`. `paper-pure` is refused by name rather than resolved — the nearest-looking option, `series`, is a different formula, so a caller who guesses would get numbers instead of an error.

## `PIO_MATRIX`

These are the first `pio_*` symbols gated on `matrix`, so `cbindgen.toml` gains the `[defines]` entry the other five optional features have and the header preamble says what the gate covers. They need `matrix` alone; the matrix Arrow tables still need `arrow` with it, which is what `pio_matrix_available()` reports.

## Verification

Six new tests: the two conventions against `x/(r²+x²)` and `1/x` computed from `pio_branches` on case9, separating exactly where the case has resistance and both positive; the shifter injecting at two buses and summing to zero under `series`/`matpower` and vanishing under `reactance-only`; a zero-reactance branch leaving a column gap that both `branch_rows` and `skipped_rows` report; an infinite reactance refused with the coded message rather than read as an open branch; the token table including the `paper-pure` refusal; and the NULL-handle path on all four.

`powerio-capi` green at default features (90), at `--features matrix`, and at full features (121). `cargo fmt --check` clean; `clippy --all-targets -D warnings` clean at default, `matrix`, full, and `--no-default-features`. `scripts/capi-header-parity.sh` and `scripts/capi-header-regen.sh` clean against the regenerated header.

## Additive

No existing symbol changes and `PIO_ABI_VERSION` stays 5.

The Julia companion is eigenergy/PowerIO.jl#114.

---

## 1.0 architecture audit

This public incidence-parts shape is superseded by #400 and ABI v6 #413. Reuse fill guards, checked conversions, mappings, and bulk extraction tests. The public equations must include `p_branch = -Bf * va + b .* shift`.


samtalki:
Superseded by the v1 architecture stack (#426): the DC matrix surface moved to the DcOperators design in #429.

samtalki:
Correction: the stack reference above is stale (#426 was renumbered). The DC matrix surface lives in the DcOperators design in #429; the fill guards, mapping code, and bulk extraction tests are carried by the ABI v6 surface in #443.

PR #404, opened 2026-08-23
mcp: expose the explicit multiconductor-to-balanced transformation

### LINKED ISSUE #398: MCP: expose the multiconductor-to-balanced transformation
## Problem

PowerIO already implements an explicit `MulticonductorNetwork` to `BalancedNetwork` transformation, but the MCP server does not expose it. PowerMCP therefore has to reject the input or call a language API outside the MCP path.

The transformation is lossy and must remain explicit. PowerIO owns its blockers, assumptions, diagnostics, stable input and output `ElementId` mappings, and descriptive history.

## Required 1.0 behavior

Expose one MCP transformation with two explicit operations:

- inspect whether the input can be transformed and return blockers, assumptions, approximations, and fields that the result cannot represent;
- construct the `BalancedNetwork` when no blocker remains.

MCP accepts `PioModule<MulticonductorNetwork>` and returns `PioModule<BalancedNetwork>`. The result preserves applicable common module records and source ownership and records the chosen base power and every assumption in the typed result and descriptive history. `TimeSeries<T>` and `ScenarioSet<T>` remain typed values; there is no repeated-values common field or public repair object. The transformation does not route through `.pio.json`.

Rust, Python, Julia, and the CLI expose the same PowerIO transformation. Parse, summary, matrix construction, conversion, and state selection never apply it implicitly.

## Completion tests

- successful inspection and transformation over MCP stdio;
- structured refusal for unsupported transformers, switches, and conductor arrangements;
- exact agreement with the Rust transformation and language bindings;
- stable input and output element mappings plus descriptive history;
- the returned module is accepted by summary, matrix, conversion, writing, and diagnostic operations;
- native multiconductor analysis remains available when the consumer accepts `MulticonductorNetwork`.

## Ownership

Primary implementation branch: `agent/v1-mcp-surface`. It advances this issue. Close only after Julia and PowerMCP parity passes. Related issues: #232, #261, and #397.

samtalki:
Closes #398. Stacked on #403 → #402 → #401.

## The gap

Python has exposed `Package.multiconductor_to_balanced_preflight` and `Package.lower_multiconductor_to_balanced` since 0.9.0. Over MCP a client could transport a multiconductor `.pio.json` package and had no way to ask for either, so a balanced solver integration had to call the Python API locally or refuse the package with no MCP-native path forward.

powerio should own the pass because it owns the assumptions, the structured diagnostics, the source mapping and the lowering history. PowerMCP refuses implicit lowering deliberately ([Power-Agent/PowerMCP#57](https://github.com/Power-Agent/PowerMCP/pull/57)), and so does this.

## The tool

```text
lower(
  package_json | path | content,
  to_model_kind="balanced",
  mode="preflight" | "apply",
  options={"base_mva": 100.0},
)
```

`to_model_kind` and `mode` are named arguments rather than two tools, so a future typed pass is an added value rather than a new verb.

**`preflight`** reports `status`, `ready`, `assumptions`, `approximations`, the full diagnostics and the `blockers` subset, and produces no model. Neither `package_json` nor `applied` appears in that response, so there is nothing to mistake for one.

**`apply`** runs the same preflight — the test asserts the two `readiness` blocks are equal — and then:

- On a blocker: `applied: false`, the blockers, and **no** `package_json` key. A structured refusal rather than a raised string, because the blockers *are* the answer and an exception message would carry one line of them. The absent key means a refusal cannot be read as "applied with an empty result".
- Otherwise: the derived balanced `package_json` with its `validation`, `diagnostics`, `origin` and `lowering_history`.

Both modes report the selected `base_mva` and sequence transform.

## `options` refuses what it cannot honor

`options` accepts `base_mva` alone. A caller who passes `convention` is told that one transform exists and the result reports it; any other key is named back. A policy-bearing lossy pass that silently drops a policy argument is the exact failure this tool exists to prevent.

## Nothing else lowers implicitly

`parse`, `summary`, `matrix`, `convert`, `save` and package readback all keep a multiconductor payload multiconductor — `test_no_other_verb_lowers_implicitly` pins that, and native multiconductor analysis stays the preferred path for a consumer that supports it.

## Acceptance criteria

| From #398 | Where |
|---|---|
| Transport tests: preflight success, structured refusal, successful application | `test_mcp_transport.py`, three new stdio tests |
| Applied package accepted by `matrix`, `convert`, `save`, `summary`, `diagnostics` | `test_lower_apply_returns_a_derived_balanced_package` |
| Python and MCP lowering equivalent | `test_mcp_and_python_lowering_produce_the_same_derived_package` (exact JSON equality) |
| `LOWER.MULTI_TO_BALANCED.*` codes preserved | untouched — the pass is called, not reimplemented |
| Convention and `base_mva` visible | `options` in every response, plus `lowering_history[-1].options` |
| Blockers stay explicit | `test_lower_apply_refuses_a_blocker_as_data` |
| No existing response changes shape | one new tool, no edits to any existing payload |

Fixtures: `fourwire_linecode.dss` is three-phase throughout and lowers clean; `switch.dss` carries a closed switch the pass will not project (`LOWER.MULTI_TO_BALANCED.UNSUPPORTED_CLOSED_SWITCH`).

## Verification

`python/tests/test_mcp.py` 47 passed, `test_mcp_transport.py` 7 passed, `test_package.py` green; `ruff check --no-fix .`, `mypy python/powerio`, and `mypy.stubtest powerio` against the checked-in allowlist all clean. The full `python/tests` run is 204 passed / 1 failed, and that one failure is `test_to_networkx_attrs_and_status_filter` needing the optional `networkx` extra, which this sandbox does not have and CI installs.

---

## 1.0 architecture audit

Preserve preflight, explicit application, diagnostics, option validation, and stdio tests. Port them to the typed `PioModule<MulticonductorNetwork>` to `PioModule<BalancedNetwork>` operation after #408 and #409. Use transformation terminology and do not merge this branch through the superseded ABI PRs.


samtalki:
Superseded by the v1 architecture stack (#426): the multiconductor surface moved to the instance and matrix designs in #428/#431.

samtalki:
Correction: the stack reference above is stale (#426 was renumbered). The preflight, apply, transformation, diagnostics, option validation, and stdio behavior are ported to the PioModule MCP surface in #442; the instance and matrix designs are #428 and #431.

PR #405, opened 2026-08-23
Address the review findings on #401–#404

samtalki:
Stacked on #404 → #403 → #402 → #401. An `xhigh` review of those four PRs' own commits (`d6a52e8`, `b58e44d`, `3c63c45`, `19e5c97`; the `main` merge in #401 was excluded as already-reviewed upstream code) produced twelve findings. Ten were applied in the initial pass. A final review resolved one deferred item and found one additional optimization; the last deferred item remains below.

## Correctness

**`Auto` no longer overrides the dense footprint veto** — *#401, `powerio-matrix/src/matrix/sensitivity.rs`*

#401 sends an `Auto` build to dense whenever the sparse path refuses the case's susceptances. That is right when the shape check chose sparse for the reduced dimension, and wrong when it chose sparse because the predicted dense footprint exceeds `AUTO_DENSE_MEMORY_BUDGET`. The fallback answers a case one path can produce; it cannot answer one neither can.

Concretely: 2 buses, 17 000 parallel branches, one with `x < 0`. `nr = 1 <= 64` passes the dimension test, `m*m*8 ≈ 2.3 GB` trips the footprint veto so the shape check returns `Sparse`, `ensure_sparse_solver_eligible` errs on the negative susceptance, and `resolve_solver` hands it to a dense builder that aborts on the allocation the veto exists to refuse. Before #401 the 8192 ceiling kept the dense fallback to cases the footprint guard had already cleared, so lowering the threshold to 64 is what exposed it.

The fallback is now conditioned on the footprint fitting; such a case keeps the sparse refusal, which names `solver=dense` for a caller who can afford it. `dense_footprint_fits` is extracted as the single predicate both the shape check and the fallback consult. Regression test: `auto_keeps_the_sparse_refusal_when_the_dense_footprint_is_vetoed`.

**One DC convention parser** — *#403, `powerio/src/dc.rs`, `powerio-py`, `powerio-capi`*

#403 wrote a second token table at the C boundary, normalizing by stripping every non-alphanumeric where `powerio-py` strips only `-` and `_`. So `pio_branch_susceptance(net, "series impedance", …)` resolved while `ptdf(convention="series impedance")` raised — under a header stating it takes "the tokens Python and the CLI accept", and a test named `the_convention_token_is_the_one_python_and_the_cli_accept`. An empty token also resolved silently to the default, in a function whose `paper-pure` arm exists precisely so a guessed convention errors instead of returning numbers.

`DcConvention::from_token` in the core is now the one parser; Python and C both call it. Python's behavior is unchanged (its `normalize` was already exactly this rule); the C surface tightens to match, and an empty token is refused. NULL still means default.

**MCP lowering readiness reads the pass's own rule** — *#404, `python/powerio/mcp/server.py`*

`MulticonductorToBalancedReadiness::is_ready()` is `status <= Info`, and `lower_multiconductor_to_balanced` refuses on exactly that — so a `warning` record blocks the pass. Deriving readiness from `error`/`fatal` alone reported `ready: true` with an empty `blockers` list beside `status: "warning"`, self-contradictory in one payload; `mode="apply"` then raised a `ValueError` carrying the first diagnostic's message — the one-line refusal that `applied: false` with structured blockers was built to replace. The registry already defines `LOWER.MULTI_TO_BALANCED.BALANCED_VALUE_DOMAIN` and `.DROPPED_LOAD_VOLTAGE_MODEL` at warning severity.

**`-1` keeps one meaning** — *#403, `powerio-capi/src/lib.rs`*

`pio_incidence_branch_rows` / `pio_incidence_skipped_rows` mapped a failed `usize → i64` cast to `-1` *inside the output array*, colliding with the value the same functions document as the whole-call refusal. The cast failure is a coded refusal now.

**Missing `lower` compatibility callable** — *#404*

`docs/src/python.md` documents `pre = lower(path=…, mode="preflight")` three lines below a block calling `summary(...)`/`matrix(...)` the same way, and every other advertised tool has a module-level callable. `from powerio.mcp.server import lower` raised `ImportError`. Added, with a test.

## Cleanup

- `base_mva` finiteness written once (`math.isfinite(...) and > 0`) instead of three overlapping conditions whose NaN arm was dead — `not (nan > 0.0)` is already `True`.
- The `# The tool text arguments that can carry JSON …` comment moved back above the `@mcp.tool` block it documents; #404 had orphaned it onto `_lowering_options`.
- Two unclosed code spans in the `lower` tool description (`` `mode="preflight" ``, `` `mode="apply" ``). A markdown-rendering client swallows the following prose into a code span — in the one text a model reads to decide how to call the tool.
- The four DC extractors document their per-call `build_incidence`, as #402's `pio_solver_index_json` already did. Header regenerated.
- `CHANGELOG.md` and `docs/src/matrices.md`: #401's bullet listed the 2 GiB veto as a reason `Auto` takes *dense*; it is a reason it takes *sparse*. Restated.

## Final review follow-up

- **Resolved: `pio_solver_index_json` built all nine solver row tables and discarded the rows.** `BalancedNetwork::to_normalized_solver_table_index` now runs the same normalization and star lowering pass but builds only the identity and source mapping vectors. The full and index-only paths share one index builder, with a regression test proving identical output.

- **Sparse sensitivity hot loops repeated reference-set searches.** Reduced branch endpoints are now computed once per build instead of twice per branch for every PTDF and LODF entry. Packed LODF batches also clear only columns that will be solved.

## Still not acted on

- **`ensure_sparse_solver_eligible` runs twice per sparse `Auto` build** — once in `resolve_solver` for its boolean, once inside `sparse_ptdf_lodf_entries`. Removing it means threading the verdict out of `resolve_solver`; an O(m) float scan is negligible against a sparse Cholesky.

## Verification

`cargo fmt --all --check`, `scripts/ci-clippy.sh workspace`, `scripts/capi-header-parity.sh` and `scripts/capi-header-regen.sh` all clean. Rust tests green including `powerio-capi` at full features (121). Python extension **rebuilt** (`powerio/src/dc.rs` and `powerio-py` both changed, so the installed one was stale) then `ruff`, `mypy`, and `pytest python/tests` — 206 passed; the single failure is the pre-existing `networkx`-not-installed skip that CI's `.[all,mcp]` install covers. MCP suites 56 passed including stdio transport.

Cross-repo: rebuilt the full-feature C ABI and re-ran the PowerIO.jl stack against it — #116 2116 passed, #117 2178 passed, so the tightened convention parser and the `-1` change don't move the Julia binding.

---

## 1.0 architecture audit

Treat this as a salvage branch, not a mergeable unit. Retain corrected PowerModels signs, token unification, allocation reductions, checked conversions, C ownership fixes, and MCP readiness work in their owning 1.0 branches. Correct every branch-flow test and document to `p_branch = -Bf * va + b .* shift`.


samtalki:
## Simplification pass on the whole stack (`a0a552b`)

A quality-only pass over the end-to-end diff of #401→#405 against `main` — reuse, simplification, efficiency, altitude; no correctness hunting. Four independent reviewers, nine deduped findings, **eight applied, one skipped**. `powerio-matrix/` has zero hunks: nothing from `b66e1e5` was touched.

### Applied

**One DC convention parser, reaching the CLI too.** #405 already made `DcConvention::from_token` the single parser for Python and C, but `powerio-cli` kept a clap `ValueEnum` listing the same three spellings a fourth time. So `--convention mp`, `--convention SERIES` and `--convention series_impedance` were refused on the command line and accepted everywhere else, and `--convention paper-pure` printed clap's generic possible-values list instead of the sentence naming the successor formula. `from_token` is now `impl FromStr for DcConvention` (matching `powerio-prob`'s `Units`), which makes it a clap `value_parser`; `DcConvArg` and its `From` impl are deleted. Verified on the built binary — all four previously-rejected tokens now resolve, and:

```
error: invalid value 'paper-pure' for '--convention <CONVENTION>': convention 'paper-pure' is now
'reactance-only'; it is no longer the default, and 'series' is a different formula (b = x/(r^2 + x^2))
```

The trade-off: clap no longer knows the values, so `--convention` does not print `[possible values: …]` while `--units`/`--solver` still do. One `CONVENTION_HELP` const names them for both subcommands, replacing the doc comment that was duplicated per field. No shell-completion regression — the repo generates none.

**`pio_scenario_ids` runs through `finish_fill`.** It is where that helper's shape came from and it kept a hand-written copy: `catch_unwind`, the `isize` conversion, both `copy_to_buf` arms. The gate widens from `matrix` to `any(matrix, gridfm)` rather than being dropped — `gridfm` enables `dep:powerio-matrix` but not capi's own `matrix` feature, and removing the gate outright fails `capi-no-default` with `error: function finish_fill is never used` under `-D dead-code`. One user-visible string moves: `"scenario count exceeds isize"` → `"count exceeds isize"`. Same `BIND_CAPI_INDEX_OUT_OF_RANGE` code, path unreachable (a `Vec` cannot exceed `isize::MAX`), and the old string appears nowhere in the repo.

**`parts_f64` and `parts_i64` were the same function twice.** `parts_of<T: Copy + Default>` follows the extractor's own `out` pointer, so both element types share it without a turbofish.

**The four DC extractor docs state the bound that makes the size query skippable.** Each already said the size query and the fill are two full incidence builds; none said `m <= pio_n_branches` bounds every vector in the family and `n == pio_n_buses` sizes the per-bus one *exactly*, which is what lets a caller drop the query rather than merely amortize it. Traced through `incidence_parts_of` → `view(net)` and checked on the 3-winding star fixture, the shape that could have broken it: `n_branches=12 n_buses=10 → b=12, rows=12, skipped=0, p_shift=10`. Tight, never exceeded. Header regenerated.

**MCP.** `_lower_choice` normalizes both `lower` enum arguments against one rule instead of two copies of the same eight lines (messages proven byte-identical across six bad inputs; `""` still defaults, `"  APPLY "` still normalizes). `_lower_impl` reads `validation` and `diagnostics` off the document it already parsed instead of calling `.validation()`/`.diagnostics()`, which each re-serialize the same package on the Rust side and re-parse it here — three passes over a document that runs to megabytes where one does; equality checked at runtime against a **non-empty** diagnostics array. `_BLOCKING_SEVERITIES` is one tuple shared with `_package_diagnostic_messages`, and `_lower_blockers` drops an `isinstance` test its input already passed.

Also: the changelog now records that `--convention` accepts the library's tokens, which is user-visible and was the one thing the validator would not have merged without.

### Skipped

**Single-pass serialization for `pio_solver_index_json`.** Proposed as byte-identical; it is not, and cannot be as specified. `serde_json`'s lock entry has no `indexmap`, so `Map` is a `BTreeMap` and `json!` sorts keys at every level, while a derive emits declaration order — `SolverTableIndex`'s 13 fields are not declared alphabetically. Built both serializers and diffed: same 532 bytes, semantically equal, different key order. Byte identity needs the struct's fields reordered or a 13-field mirror struct; that is a maintainer's call, so the rewrite was reverted.

### Verification

`cargo fmt --all --check`, `scripts/ci-clippy.sh all` (7 configs, `-D warnings`), `capi-header-parity.sh`, `capi-header-regen.sh` all clean. `cargo test --workspace` 1475 passed / 0 failed across 61 binaries; `powerio-capi` at full features 121 passed. `RUSTDOCFLAGS="-D warnings" cargo doc` clean over 8 crates. Extension rebuilt (`dc.rs` and `powerio-py` both changed), then `ruff`, `mypy` and `pytest python/tests` — **213 passed, 2 skipped**, MCP suites included.

---

samtalki:
Superseded by the v1 architecture stack (#426). The sensitivity allocation findings this addressed are carried by #432; the C ABI and MCP surfaces from #402-#404 were replaced by the v1 module layout.

samtalki:
Correction: the stack reference above is stale (#426 was renumbered). The sensitivity allocation findings are carried by #432, the PowerModels sign conventions by #429 and #443, and the MCP readiness fixes by #442.

PR #475, opened 2026-09-03
Decode the .pwb bus record family with flag bit 2 clear

samtalki:
A local corpus of twenty PowerWorld saves was refused outright: every one of
them writes bus record flag words with bit 2 clear (0x3002, 0x3003, 0x3022,
0x3062, 0x3162, 0x3163), and the reader required that bit set. Bit 2's field is
not in the decoded head, so it joins the presence set; it stays in the record
family key so a bit 2 clear table cannot chain with a bit 2 set one. Bit 5, the
tail family bit, is clear over one of the twenty bus tables and set over the
rest, so this vintage writes both tails. Header constant 537 also turns out to
admit the 425 era generator record, so it now tries both shapes in sequence as
508 already does. `FORMAT.md` records the flag word census, the accept rule, and
what each claim was checked against.

Two decode gains follow. The 425 era generator record carries the 2021 era
status tail on the writers that emit it (zero byte at block +32, status byte at
+33, GenRMPCT at +34); the reader reads the status only when every record of the
accepted table carries the tail, so a table without it still reports the gap.
And a `.pwd` display whose header carries a canvas title was refused as
unrecognized, because the title shifts the per file stamp past offset 22: a 1.0
reader defect, fixed and pinned by a test that inserts a title into the vendored
display and requires the same substations back.

All twenty binaries decode and match the MATPOWER and PSS/E exports of the same
cases element for element on every quantity the reader decodes. That check ships
as an opt in test: `POWERIO_PWB_CASES` names a local directory, unset skips, and
pairing is by bus identity and solved state rather than by file name. Nothing is
vendored.

A new `.pwb` vintage is additive under the frozen 1.0 surface; the `.pwd` fix is
a defect fix and belongs in 1.0.

### DIFF
--- a/powerio-tx/src/format/powerworld/FORMAT.md
+++ b/powerio-tx/src/format/powerworld/FORMAT.md
@@ -151,14 +151,16 @@ quantum, dispatch and branch values likewise. `powerio/tests/`
 
 ## The .pwb binary format (decode evidence)
 
-Established by differential analysis of three lawfully obtained files, no
+Established by differential analysis of lawfully obtained files, no
 PowerWorld software involved: ACTIVSg200.pwb (Simulator 20 era, June 2018,
 same snapshot as the vendored aux), Texas2000_June2016.pwb (June 2016, same
 day as its paired aux export), and ACTIV_SG_2000_v19.pwb (April 2017, validated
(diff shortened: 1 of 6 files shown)


samtalki:
Superseded by #477, which carries the same commits from a branch whose name does not identify the local case corpus.

PR #480, opened 2026-09-03
docs(test): harden deferred OpenDSS geometry against silent defaults

BurhanAbdullah:
Superseded by #495. Both contributed geometry cases run there as ordinary regression tests in `powerio-dist/tests/electrical_readiness.rs`, with contributor credit in the PR description. #495 closes #479 when merged.

## Summary

This draft contributes a regression harness and a concrete safety contract for the geometry-normalization issue in #479.

The current reader can normalize `Line` objects that reference the deferred OpenDSS geometry family using the `Line` class factory impedance defaults, while also deriving conductor count from the balanced phase default. The result can be electrically wrong before any writer is invoked.

### Included
- Regression harness for a geometry-backed overhead line.
- Regression harness for a single-conductor SWER geometry.
- Explicit parse-time-diagnostic expectation.
- Documentation of the fail-closed contract and the required long-term canonical representation.

### Why draft

The canonical `DistLine` currently requires a concrete `linecode` and impedance matrices. A complete fix therefore needs an intentional representation for a deferred/unresolved line rather than replacing the impedance with a sentinel that could accidentally reach a solver. I have deliberately not introduced a speculative schema change in this PR.

The target invariant is:

> Unsupported geometry must remain source-retained and explicitly diagnosed; it must never silently acquire OpenDSS factory impedance or a fabricated conductor count.

Explicit `linecode=` paths remain unchanged.

Closes the implementation gap described by #479 once the deferred-line representation is agreed.

Issue #479, opened 2026-09-03
Geometry-defined lines are replaced by OpenDSS factory defaults in the normalized model, with a fabricated conductor count and nothing reported at parse time


archiechapman:
**PowerIO:** 0.10.0 (current release) · **Julia:** 1.11.3 · **Windows** · cross-checked against OpenDSS via OpenDSSDirect.jl 0.9.9

## Summary

#84 says deferred OpenDSS classes are kept in the raw object layer with explicit conversion warnings, "so unsupported data does not disappear silently". For the geometry family — `LineGeometry`, `WireData`, `CNData`, `TSData`, `LineSpacing` — that invariant does not hold. The data does not merely stay untyped: a **placeholder impedance built from OpenDSS's factory `Line` defaults** (`R1=0.058, X1=0.1206, R0=0.1784, X0=0.4047`) is substituted in its place, together with a **fabricated conductor count of 3**.

Three things make it worse than a deferral:

1. It happens in the reader's normalization, so it is visible from `linecodes(net)` **before any writer runs**, and `diagnostics(m)` on the parsed module is **empty**. The warnings exist only as writer-emitted findings.
2. `"bmopf"` and `"pmd"` both inherit it. `"opendss"` does not, because it re-emits the retained source records — so a `dss → dss` round-trip looks clean and hides it.
3. The result is a network that parses, validates and solves. "Does it solve?" is not a check that catches this.

I am raising this separately from #84 because it is not a request for geometry support — it is that the deferral is not currently silent-safe. Four of the five suggestions in §5 do not depend on implementing Carson.

Frederik has confirmed the downstream half of this on the BMOPFTools side; this issue is the PowerIO-side scoping, which turned out to be wider than the BMOPF writer we originally suspected.

## 1. Reproducer

Two files and a script, all inline below. One circuit, eight lines, each defining its impedance a different way.

<details>
<summary><code>repro.dss</code></summary>

```
! One circuit, eight lines, each defining its impedance a different way.
! Conductors are Kersting 556,500 26/7 ACSR (phase) and 4/0 6/1 ACSR (neutral).
clear
new circuit.repro basekv=4.16 phases=3 bus1=sourcebus
set defaultbasefreq=60
set earthmodel=Carson

new wiredata.acsr556 rac=0.1859 gmr=0.0313 radius=0.4635 runits=mi gmrunits=ft radunits=in normamps=730
[...]

Issue #84, opened 2026-06-10
powerio-dist: typed support for deferred OpenDSS classes


samtalki:
The static OpenDSS profile converts supported equipment into `MulticonductorNetwork`. Unsupported objects remain in retained source data and receive diagnostics before canonical conversion or calculation.

Typed support in the current source includes:

- `Circuit` / `Vsource`, `Line`, `LineCode`, and switch-like lines.
- `Load`, `Generator`, and `Capacitor` equipment.
- `Transformer` winding data, including n-winding pair reactances, ratings, taps, neutral impedances, and per-phase bounds. Matrix builders check whether the particular connection model is supported.
- Grounding `Reactor` objects as inductive shunts; series and impedance reactors still need broader typed support.
- `SwtControl` state snapshots and `RegControl` fixed-tap snapshots.
- `PVSystem`, `InvControl`, and `XYCurve` data through the typed inverter and control-profile records. This is a static electrical profile, not a control-sequence simulator.

Deferred work remains:

- `Storage` and its operating/control semantics.
- `LoadShape`, `GrowthShape`, `TShape`, `PriceShape`, and time-dependent evaluation of references to them.
- `Fault`, `Fuse`, `Recloser`, `Relay`, and protection sequencing.
- `CapControl` and dynamic regulator behavior beyond the fixed-tap snapshot.
- `XfmrCode` as a first-class parameter source, and calculation support for transformer/regulator arrangements beyond the implemented models.
- `LineGeometry`, `WireData`, `CNData`, `TSData`, and impedance calculation from unreduced geometry. Unresolved geometry must remain explicit and cannot acquire factory impedances or fabricated conductor counts. #495 carries the ordinary regressions and closes #479; #480 is closed as superseded with contributor credit.
- Instrumentation and reporting classes such as `EnergyMeter`, `Monitor`, and `Sensor`; their need for typed records depends on the calculation profile.

#498 adds a separate neutral-Kron projection and a documented, restricted LinDist3Flow formulation. Geometry-derived impedance calculation and the deferred classes above remain open under this issue.

New typed classes should use the OpenDSS definition-order property table and documented defaults, report target-format losses, and include an independent OpenDSS solve comparison when the class changes power-flow behavior.

samtalki:
[...]

PR #484, opened 2026-09-04
Fail closed on deferred OpenDSS geometry

samtalki:
## Why

OpenDSS lines whose impedance comes from `LineGeometry`, `LineSpacing`,
`WireData`, `CNData`, or `TSData` currently fall through to the `Line` class
factory impedance. That can turn a real geometry into plausible but wrong
electrical parameters and can turn a one-conductor SWER line into three
conductors without a parse diagnostic (#479).

This implementation is intentionally stacked on Burhan Abdullah's regression
and safety-contract PR #480. The temporary base branch is an exact mirror of
#480's head so this PR contains only the follow-up implementation commit. Merge
#480 first; this PR can then be retargeted to `main` without copying the
contributor's work.

## What changed

- Represents deferred line impedance explicitly as `DistLine.linecode: None`.
- Replays `phases`, `linecode`, `geometry`, `spacing`, and conductor/cable-list
  assignments in source order so terminal maps retain the stated conductor
  shape.
- Emits a parse-time error diagnostic when geometry impedance has not been
  calculated.
- Preserves exact same-format DSS output from retained source.
- Omits unresolved lines from semantic DSS, PMD, and BMOPF output with an error
  diagnostic instead of inventing a linecode or electrical values.
- Keeps explicit linecode and inline-impedance behavior unchanged.

## Validation

- `cargo test -p powerio-dist`
- `cargo check -p powerio-dist --all-features`
- `cargo clippy -p powerio-dist --all-targets -- -D warnings`
- `cargo fmt --all --check`
- `git diff --check`

The regression coverage includes four-conductor geometry, single-conductor
SWER, spacing plus conductor lists, parse severity, exact DSS echo, and every
semantic distribution writer.

Depends on #480. Addresses #479 without implementing the separate Carson/line
constants work tracked in #84.

### DIFF
--- a/docs/dss-geometry-safety.md
+++ b/docs/dss-geometry-safety.md
@@ -15,7 +15,8 @@ a geometry-defined line.
 
 Until Carson/geometry impedance calculation is implemented:
 
-1. Detect a `Line` that references `geometry=`, `spacing=`, or `wires=`.
+1. Detect a `Line` that references `geometry=`, `spacing=`, `wires=`,
+   `cncables=`, or `tscables=`.
 2. Do not synthesize a `DistLineCode` from OpenDSS `Line` factory defaults.
 3. Do not infer/fabricate conductor count from the balanced `phases` default.
 4. Emit a **parse-time diagnostic** identifying the deferred geometry source.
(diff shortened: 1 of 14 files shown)


BurhanAbdullah:
I traced the workspace compile failure after `DistLine.linecode` became `Option<String>`. The core direction in this PR is right; the remaining issue is propagation of the new unresolved state into downstream consumers.

The current failures are in facade code that still assumes `linecode: String`, notably `powerio/src/stored/legacy09/document.rs` and `powerio/src/transform.rs` (calls such as `net.linecode(&line.linecode)`, `eq_ignore_ascii_case(&line.linecode)`, and formatting `line.linecode` with `{}`). These should not unwrap/default `None`, because that would reintroduce the silent-fabrication hazard this PR is fixing.

Suggested invariant for each consumer:

```rust
let Some(linecode) = line.linecode.as_deref() else {
    // record/return the existing unresolved-impedance error for this semantic path
    // and do not lower/serialize/analyze this line
    continue; // or return the appropriate readiness/lowering error
};
```

Then use `linecode` for lookup/diagnostic formatting. In readiness/lowering code, `None` should make the network not electrically lowerable rather than appearing as an unknown named linecode.

I also searched the tree for remaining direct `line.linecode` assumptions. Besides the two compile-failing facade paths, `powerio-matrix/src/matrix/multiconductor.rs` and some tests still use the old mandatory-string shape on current main. Worth auditing those as part of the same API propagation so a future all-workspace build cannot accidentally accept unresolved geometry.

Happy to keep validating the propagation against #480's fail-closed contract; the important point is preserving `None` as a first-class unresolved electrical state, never coercing it to a placeholder.

PR #494, opened 2026-09-05
feat(dist): add fail-closed electrical readiness gate

BurhanAbdullah:
## Summary

- add a read-only `audit_electrical_readiness` gate for `MulticonductorNetwork`
- treat deferred OpenDSS `geometry=`, `spacing=`, `wires=`, `cncables=`, and `tscables=` as hard readiness blockers
- catch unresolved endpoints/linecodes, invalid lengths, duplicate IDs, malformed impedance matrices, and non-finite matrix values
- add active parser-to-audit regression coverage for both multi-conductor geometry and one-conductor SWER geometry
- document the intended pre-solver safety boundary

## Rationale

This addresses the safety half of #479 without attempting to implement Carson/geometry lowering. A parsed geometry-backed line may still exist in the typed model for source retention, but callers now have an explicit, non-breaking fail-closed check before numerical lowering or solving. The audit does not infer conductor count or accept OpenDSS factory impedance as evidence of electrical completeness.

## Validation

The branch contains unit tests and parser-to-audit integration tests. GitHub Actions is the authoritative CI check; no local test result is being claimed here before CI runs.

### DIFF
--- a/docs/electrical-readiness.md
+++ b/docs/electrical-readiness.md
@@ -1,0 +1,52 @@
+# Electrical readiness before analysis
+
+`powerio-dist` separates source parsing from the decision to hand a
+multiconductor network to numerical analysis. Parsing establishes a typed
+model; [`audit_electrical_readiness`](../powerio-dist/src/readiness.rs) checks
+whether that model is electrically complete enough for a solver or semantic
+lowering pass.
+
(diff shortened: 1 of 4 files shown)


samtalki:
Thank you for contribution @BurhanAbdullah , running CI now :)

Issue #511, opened 2026-09-15
Extend LinDist3Flow to accept meshed networks and broader component approximations


frederikgeth:
edit: there's essentially a couple of formulation variants between following the literature related to LinDist3Flow rigorously vs more experimental, accept-every-bmopf-case and do your best attempt at constructing an approximate linear model. I'm not sure at this stage who the general audience for this is yet, so happy to delete the options we don't use or don't want to use.

--

## Planned contribution

I plan to extend PowerIO’s LinDist3Flow implementation to accept a broader range of distribution networks for exploratory optimisation studies, including meshed networks and networks requiring approximate component representations.

The implementation would draw on [FormulationLab.jl PR #1](https://github.com/frederikgeth/FormulationLab.jl/pull/1) and its underlying component formulations, with one deliberate difference: **this contribution will not add the proposed angle-drop or loop-consistency equations.**

The aim is to keep electrical preparation and formulation assembly in PowerIO, making the capability available to downstream applications such as Tellegen.

## Intended modelling scope

LinDist3Flow should be a broadly usable, **use-at-your-own-risk linear optimisation model**: affine electrical equations with linear or conic limits, rather than a universal AC-feasibility certificate.

I favour accepting networks through documented approximations even where there is no settled scientific approach covering every component combination. Comprehensive AC validation should not be a prerequisite for constructing and using the approximate model.

Results should clearly identify assumptions, component transformations, inferred values and omitted constraints. Successful optimisation means feasibility under the constructed model—not necessarily physical feasibility, accurate branch flows, uniqueness, or a certified objective bound.

## Accept meshed networks without additional loop equations

The current implementation rejects conductor cycles and uses source-rooted forest assumptions in topology orientation and reference propagation.

The proposed change would:

- Accept cycles and parallel lines while retaining every in-service branch.
- Replace forest-only traversal with cycle-safe topology processing.
- Assign consistent branch/conductor orientations and preserve signed flow identities.
- Retain nodal active/reactive power balances and the existing squared-voltage drop equations.
- Make reference construction terminate deterministically on cyclic graphs, with diagnostics for unresolved or conflicting references.
- Preserve source-coverage and connectivity checks where required by the constructed model.

**No additional angle-drop, angle-recovery or cycle-consistency constraints would be introduced.** Nor would this contribution add circulating-flow penalties, regularisation or automatic radialisation.

[...]

PR #530, opened 2026-09-17
Accept meshed topology in LinDist3Flow

frederikgeth:
## Summary

- accept LinDist3Flow conductor cycles and parallel lines instead of rejecting non-radial topology
- orient conductors deterministically from physical sources and propagate reference voltages without cycle recursion
- retain every line's balance and voltage-drop equations while marking meshed formulations explicitly
- report the omitted loop/angle-consistency equations as a structured approximation diagnostic

## Stack

This is PR 1 of 3 and targets `main`. #531 adds component preparation, followed by fixed-dispatch results in #532.

Part of #511.

## Validation

- `cargo test`
- `bash scripts/ci-clippy.sh`
- `cargo fmt --all --check`


samtalki:
v0.11.3 is tagged (190336b7), so this stack lands in 0.11.4. A rebase onto main needs three things beyond the usual conflicts:

- The CHANGELOG bullets go under a new `## Unreleased` section above `## 0.11.3` (the heading these branches added to was renamed by the release; a docs PR opening that section is on its way).
- `docs/schema/pio-ir/2/0.11.3/schema.json` is now the frozen catalog of the released 0.11.3 types, so the LinDist3Flow fixed-dispatch types need their own catalog directory (the next release's) instead of a second copy of that file; `IR_SCHEMA_ID`, `frozen_schemas.rs`, `ir_reference.rs` and `scripts/check-value-types.sh` follow that path.
- The value types register next to the contingency ones added in 0.11.3 in `powerio/src/lib.rs`, `powerio/src/value.rs`, `powerio-py/src/lib.rs` and the Python stubs, which is where the textual conflicts will be.

Happy to do the rebase of the three branches if that is easier; say so and I will push it to the same branches.

frederikgeth:
Go for it.

